PR #9, opened 2021-12-14
[NA-245] - remove UserGet Schema base_token field

bjornaer:
commented out base token while checking if it would be a breaking change

also commented test cases that broke due to commented function in codebase

### DIFF
--- a/pipeline/schemas/user.py
+++ b/pipeline/schemas/user.py
@@ -15,14 +15,17 @@ class UserBase(AvatarHolder):
     company: Optional[str]
     job_title: Optional[str]
 
+
 class UserGet(UserBase):
     id: str
     oauth_provider: Optional[str]
     verified: Optional[bool]
     subscribed: Optional[bool]
 
+
 class EnrichedUserGet(UserGet):
     base_token: TokenGet
+
 
 class UserGetDetailed(UserGet):
     tokens: List[TokenGet] = []
--- a/pipeline/schemas/user.py
+++ b/pipeline/schemas/user.py
@@ -18,7 +18,7 @@ class UserBase(AvatarHolder):
 
 class UserGet(UserBase):
     id: str
-    base_token: TokenGet
+    # base_token: TokenGet
     oauth_provider: Optional[str]
     verified: Optional[bool]
     subscribed: Optional[bool]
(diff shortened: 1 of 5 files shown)


alexpearce:
What's the commented-out function? Do you think it's worth re-adding, given we have to drop several tests? Maybe @paulcjh can comment on whether the code should be uncommented.

bjornaer:
> What's the commented-out function? Do you think it's worth re-adding, given we have to drop several tests? Maybe @paulcjh can comment on whether the code should be uncommented.

I have not done so because I lack context as to why it was commented in first place -- so i adjusted tests to just stop breaking

paulcjh:
> > What's the commented-out function? Do you think it's worth re-adding, given we have to drop several tests? Maybe @paulcjh can comment on whether the code should be uncommented.
> 
> I have not done so because I lack context as to why it was commented in first place -- so i adjusted tests to just stop breaking

The tests are super out of date, I've had to omit them from my personal workflow for the moment because of how frequently I've been changing the schemas. For the moment don't worry about them. We'll sort this out in a future PR.

pipeline/schemas/user.py
```diff
@@ -18,12 +18,15 @@ class UserBase(AvatarHolder):
...
 class UserGet(UserBase):
     id: str
-    base_token: TokenGet
     oauth_provider: Optional[str]
     verified: Optional[bool]
     subscribed: Optional[bool]
 
 
+class EnrichedUserGet(UserGet):
+    base_token: TokenGet
```
alexpearce, line 27: Unfortunately the [naming conventions](https://github.com/neuro-ai-dev/pipeline_schemas#naming-conventions) didn't survive the schema migration but `UserGetEnriched` would be more consistent with other 'extended' names. I wouldn't be against something more explicit here e.g. `UserGetWithBaseToken`.

alexpearce (approved):
LGTM, thank you! Just a minor suggestion from me for your consideration.

pipeline/schemas/user.py
bjornaer: I thought about making it specific, but then erinched felt it delivers the idea and allows for more data to be added in the future, I switch around the wording!

PR #10, opened 2021-12-14
[NO-TICKET] add automated checks, change code to comply

bjornaer:
added:

- `black` linting
- `isort` for imports
- `flake8` to check code formatting

- and `pre-commit` to enable pre commit hooks for those checks

after starting to fix some errors I went down a rabbit hole and made a lot of changes

So this PR is now a Pipeline design proposal with tests and examples.

Main thing to look at is `pipeline/objects/pipeline_v2.py`


pipeline/__init__.py
```diff
@@ -1,11 +1,8 @@
...
-from typing import List
+from pipeline.objects.pipeline_v2 import PipelineV2
 
-from dill import loads, dumps
-
-# from pipeline import logging
+__all__ = ["PipelineV2"]
 
 
 CACHE_DIR = os.getenv("PIPELINE_CACHE_DIR", "./cache")
```
alexpearce, line 8: An aside while we're overhauling stuff: we should consider conforming to the [XDG base directory specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) here. Specifically, using `$XDG_CACHE_HOME` as the base directory rather than `.`, falling back to `$HOME/.cache` if that environment variable is empty or not set.

pipeline/api/__init__.py
```diff
@@ -47,7 +43,8 @@ def authenticate(token: str, url: str = PIPELINE_API_URL):
     PIPELINE_API_TOKEN = token
     global PIPELINE_API_URL
     PIPELINE_API_URL = url
-    # TODO: Change this url to an actual auth one, not status which just shows if the API is alive.
+    # TODO: Change this url to an actual auth one,
+    # not status which just shows if the API is alive.
```
alexpearce, line 47: I'd say this comment can be removed; `GET /v2/users/me` _is_ an endpoint which requires authentication. I think this comment is a remnant from when the check hit `GET /v2/status`, which does not require authentication.

pyproject.toml
```diff
@@ -30,8 +32,30 @@ pydantic = "^1.8.2"
...
+    \.mypy_cache
+  )/
+)
+'''
+[tool.isort]
+multi_line_output = 3
+include_trailing_comma = true
+force_grid_wrap = 0
+use_parentheses = true
+line_length = 79
```
alexpearce, line 61: Can we be more succinct here and use [isort's Black compatibility mode](https://pycqa.github.io/isort/docs/configuration/black_compatibility.html) instead?

setup.cfg
```diff
@@ -0,0 +1,2 @@
+[flake8]
+extend-ignore = E203
```
alexpearce, line 2: Can the configuration here be included in the `pyproject.toml` instead, or does flake8 not support that yet?

alexpearce:
I've left a few nit-picking comments inline. The linting additions are great; the only addition I'd like to see there is some instructions for installing the pre-commit hooks when starting development, perhaps something in the README, a "development" section or something?

---

On the more general change, it looks like you've gone from the original API:

```python
@function
def add(x, y):
    return x + y


with Pipeline("name") as pipeline:
    # …
```

to a new API:

```python
pipeline = Pipeline("name")


@pipeline.stage
def add(x, y):
    return x + y
```

This is a big change so warrants big discussion! 😄 

A bit of context first: any considerations we make should be made with a view to the [user stories](https://neuro-ai.atlassian.net/wiki/spaces/NA/pages/48660509/User+stories), which try to capture what types of things we expect users will want to do. Any API design should:

- Be flexible enough to support the various stories.
- Be nice to use. 
- Have an eye to future extensions.

(I'm not saying the current API does or does meet these, but that's the goal!)

From a birds-eye view it seems that the new design does not support re-use of functions in multiple pipelines. By changing the decorator to `pipeline_instance.stage` the user must instantiate a `Pipeline` before defining a function/stage which can be used within a pipeline, but then that function/stage is tightly bound to _that_ pipeline instance. This would prevent us from e.g. providing common utility stages within the `pipeline` module that users could plug into their own pipelines.

There's no explicit chaining of functions within a pipeline. It looks like function order is defined as the order of definition, such that the output of the first `pipeline.stage` is used as input to the next `pipeline.stage`. This is quite restrictive, as a user must define things in order, but also a bit unintuitive for me as I've not seen a decorator being used with an 'ordering' like this before, so it's unexpected. It also prevents functions being reused in multiple places in the pipeline. Can you reproduce the [pipeline in the `basic_maths` example](https://github.com/neuro-ai-dev/pipeline/blob/913f24529ab82668a7eee22a6cd8a6ca9a9fddcf/examples/basic_maths/basic_maths.py#L22-L35) with the new API? My understanding is that the new API can only construct a linear sequence of functions, whereas we want to be able to construct a [DAG](https://en.wikipedia.org/wiki/Directed_acyclic_graph).

That's probably enough feedback for now, as addressing the above points would be quite a bit of work. I guess the most interesting question for you is: what motivated you to design a new API? What aspects of the new design do you feel improve on the existing design?

setup.cfg
bjornaer: flake8 does not support it yet, afaik, if you have any resource that says otherwise, please I'd be more than happy to move this

alexpearce: You're [right](https://flake8.pycqa.org/en/latest/user/configuration.html#configuration-locations) 😢

PR #12, opened 2021-12-21
deploy automatically to test-pypi on push and pypi on tagged push

bjornaer:


alexpearce:
A couple of general questions:

1. Do we need to worry about builds for different platforms and Python versions? AFAIK the package shouldn't depend on the platform as we don't build any binaries, but I'm not sure.
2. Is it possible to test that the package built by the CI is installable? Perhaps check the tests pass when run from the package? (I mean, in a clean environment without access to the source.)

bjornaer:
> A couple of general questions:
> 
> 1. Do we need to worry about builds for different platforms and Python versions? AFAIK the package shouldn't depend on the platform as we don't build any binaries, but I'm not sure.
> 2. Is it possible to test that the package built by the CI is installable? Perhaps check the tests pass when run from the package? (I mean, in a clean environment without access to the source.)

1. We should not worry about platforms I think, but we should be wary of python versions as we develop we might be using version specific things that might not be retrocompatible
2. I think there is a way to run a check in the installation, lemme investigate

.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,40 @@
+name: Publish Pipeline-ai 🐍 distributions 📦 to PyPI and TestPyPI
+
+on: push
+
+jobs:
+  build-and-publish:
+    name: Build and publish Pipeline-ai 🐍 distributions 📦 to PyPI and TestPyPI
+    runs-on: ubuntu-18.04
```
alexpearce, line 8: Is there a particular reason to pick 18.04? If not I'd be tempted to use `ubuntu-latest` instead.

.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,40 @@
...
+        python -m
+        build
+        --sdist
+        --wheel
+        --outdir dist/
+        .
+    - name: Publish distribution 📦 to Test PyPI
+      uses: pypa/gh-action-pypi-publish@master
+      with:
+        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
```
alexpearce, line 32: Could you please add this secret to the development vault?

.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,40 @@
...
+        repository_url: https://test.pypi.org/legacy/
+        verbose: true
+        skip_existing: true
+    - name: Publish distribution 📦 to PyPI
+      if: startsWith(github.ref, 'refs/tags')
+      uses: pypa/gh-action-pypi-publish@master
+      with:
+        password: ${{ secrets.PYPI_API_TOKEN }}
```
alexpearce, line 40: Could you please add this secret to the development vault?

.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,40 @@
...
+        --user
+    - name: Build a binary wheel and a source tarball
+      run: >-
+        python -m
+        build
+        --sdist
+        --wheel
+        --outdir dist/
+        .
+    - name: Publish distribution 📦 to Test PyPI
```
alexpearce, line 29: What version is used for these development builds? If we have two concurrent PRs will their workflows end up overwriting the same version on Test PyPI?

pyproject.toml
```diff
@@ -1,11 +1,13 @@
 [tool.poetry]
 name = "pipeline-ai"
-version = "0.0.8"
+version = "0.0.9"
 description = "Yay ml pipelines"
```
alexpearce, line 4: @paulcjh we need a better description 😆

setup.py
```diff
@@ -1,11 +1,14 @@
 import setuptools
```
alexpearce, line 1: You mentioned on Slack that only "some dev cases" require a `setup.py`. Do we depend on those? If not, how about we just delete this?

bjornaer: yeah, I think we should be fine deleting the setup.py

.github/workflows/publish-to-pypi.yml
bjornaer: I added a boolean value allowing duplicate versions to be deployed to test-pypi as the idea is to have constant deployments there, but the actual pypi will throw if this case happens

.github/workflows/publish-to-pypi.yml
bjornaer: true

.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,40 @@
+name: Publish Pipeline-ai 🐍 distributions 📦 to PyPI and TestPyPI
+
+on: push
```
alexpearce, line 3: Maybe this is obvious, but the push to real PyPI only happens if the commit ref is tag. If I push to GitHub, then create the tag and push that, will the workflow be re-run, triggering the 'real' release? I guess I'm wondering if "push" means only "commit pushes" or also "tag pushes".

bjornaer: means any push onto a branch

.github/workflows/publish-to-pypi.yml
bjornaer: removed to avoid duped versions

PR #13, opened 2021-12-23
remove setup.py and requirements.txt

bjornaer:


.gitignore
```diff
@@ -136,4 +136,5 @@ examples/examples
 */hidden.env
 local/
 pipeline-bak/
-*hidden*.env
\ No newline at end of file
+*hidden*.env
+hidden.env
```
alexpearce, line 140: Does `*hidden*.env` not match `hidden.env`?

alexpearce (approved):
LGTM! 💣

.gitignore
bjornaer: it was not ignoring hidden.env
tbh .gitignore sometimes eludes me

alexpearce: How about we change this to `*.env`? From what I've seen these are always 'local' files. If examples exist they have different extensions (`*.env.example`) which a dev can copy to `*.env`.

PR #15, opened 2022-01-05
Add pre-commit hooks

bjornaer:


pipeline/api/__init__.py
```diff
@@ -47,7 +41,8 @@ def authenticate(token: str, url: str = PIPELINE_API_URL):
     PIPELINE_API_TOKEN = token
     global PIPELINE_API_URL
     PIPELINE_API_URL = url
-    # TODO: Change this url to an actual auth one, not status which just shows if the API is alive.
+    # TODO: Change this url to an actual auth one,
+    # not status which just shows if the API is alive.
```
alexpearce, line 45: We can remove this TODO, it's done!

pipeline/api/run/__init__.py
```diff
@@ -15,13 +10,13 @@ def run_pipeline(
...
-        temp_file = io.BytesIO(python_object_to_hex(data_or_file_id).encode())
-        uploaded_data = upload_file(temp_file, "/")
-        file_id = uploaded_data.id
+    # file_id = None
+    # if isinstance(data_or_file_id, str):
+    #     file_id = data_or_file_id
+    # else:
+    #     temp_file = io.BytesIO(python_object_to_hex(data_or_file_id).encode())
+    #     uploaded_data = upload_file(temp_file, "/")
+    #     file_id = uploaded_data.id
```
alexpearce, line 19: Is this block just plain broken atm?

pipeline/api/run/__init__.py
```diff
@@ -30,11 +25,12 @@ def run_pipeline(
         pipeline_id = pipeline_id_or_schema.id
     else:
         raise Exception(
-            "Must either pass a pipeline id, or a pipeline get schema. Not object of type %s in arg 1."
-            % str(pipeline_id_or_schema)
+            "Must either pass a pipeline id, or a pipeline get schema.",
+            "Not object of type %s in arg 1." % str(pipeline_id_or_schema),
```
alexpearce, line 29: The comma here is new, so the exception message is now a tuple. Is that intentional?

pipeline/objects/__init__.py
```diff
@@ -23,12 +22,13 @@ def execute_func(*args, **kwargs):
...
                 elif hasattr(input_arg, "__pipeline_model__"):
-                    if function.__pipeline_function__.class_instance == None:
+                    if function.__pipeline_function__.class_instance is None:
                         function.__pipeline_function__.class_instance = input_arg
 
                 else:
                     raise Exception(
-                        "You can't input random variables, follow the way of the Pipeline. Got type"
+                        "You can't input random variables,",
+                        "follow the way of the Pipeline. Got type",
```
alexpearce, line 31: Same comma comment here. We could add the type to the message while we're here; looks like it should show `type(input_arg)`.

pipeline/objects/__init__.py
```diff
@@ -77,10 +77,10 @@ def __function_exe__(self, *args, **kwargs):
...
-                for model_attr in dir(created_model)
-                if callable(getattr(created_model, model_attr))
-                and model_attr[:2] != "__"
-            ]
+            # model_functions = [
+            #     model_attr
+            #     for model_attr in dir(created_model)
+            #     if callable(getattr(created_model, model_attr))
+            #     and model_attr[:2] != "__"
+            # ]
```
alexpearce, line 85: I see this is unused; maybe just delete it 💣

pipeline/objects/function.py
```diff
@@ -42,10 +32,12 @@ def __init__(
...
         self.hash = sha256(self.source.encode()).hexdigest()
 
-        # TODO: Add verification that all inputs to function have a typing annotation, except for "self"
-        if not "return" in function.__annotations__:
+        # TODO: Add verification that all inputs to function have a typing annotation,
+        # except for "self"
+        if "return" not in function.__annotations__:
             raise Exception(
-                "You must define an output type for a piepline function. e.g. def my_func(...) -> float:"
+                "You must define an output type for a piepline function.",
```
alexpearce, line 39: Comma.

pipeline/schemas/user.py
```diff
@@ -46,7 +46,8 @@ class UserUsernamePatch(Patchable):
     def validate_username(cls, value):
         if not valid_username(value):
             raise ValueError(
-                "must contain between 3-24 characters, only alphanumerics, hyphens and underscores."
+                "must contain between 3-24 characters, only alphanumerics,",
+                "hyphens and underscores.",
```
alexpearce, line 50: Commas.

alexpearce: Will omit similar comments in this file.

alexpearce (changes requested):
This a great addition! Thanks a lot. I have a few requests; the commented-out code can just go IMO.

pipeline/api/run/__init__.py
bjornaer: unused

pipeline/api/run/__init__.py
bjornaer: no, will fix all the similar ones

alexpearce: Thanks! Be aware that the concatenation will not add spaces so you'll have to add them by hand

```python
>>> (
...    "abc"
...    "xyz"
... )
'abcxyz'
```

pipeline/api/run/__init__.py
alexpearce: I guess `file_id` is unused but I could imagine `upload_file` has side effects. Are you sure it's OK to remove it?

bjornaer: it posts the file to api and returns the response

- but this relates to the sort of issue we should tackle, we can't be concerned with commenting a function in case its got side effects happening that we are not aware of!! in any case, I just uncommented and printed the var to skip the error quickly as having the code pristine is not the goal of this PR, but just adding the checks

alexpearce: > we can't be concerned with commenting a function in case its got side effects happening that we are not aware of

In this case I think there was confusion with the fact that `file_id` was unused, which led to the whole block being commented without an understanding of the underlying intent. This is something we can improve on with clearer docs a broader test coverage, rather than an intrinsic problem with the structure of the code.

Agreed in general that there's a lot we can do to make the developer experience much better here.

alexpearce (approved):

alexpearce:
Thanks a lot for the changes during review, LGTM :+1:

Is this ready to go from your end? The CI is failing but I think you said this is expected, and something to fix in a separate PR.

bjornaer:
> Thanks a lot for the changes during review, LGTM 👍
> 
> Is this ready to go from your end? The CI is failing but I think you said this is expected, and something to fix in a separate PR.

yeah, it's good to go on my end, it is failing from api_integration branch as well, before doing a merge to main again I would tackle this on a sepparate PR

PR #18, opened 2022-01-06
Max/move api in pipeline

bjornaer:


paulcjh (changes requested):
Good changes, very tidy. Only change that would be good to make here would be to rename "PipelineAPI" to "PipelineCloud" just to keep consistency with https://pipeline.ai

PR #20, opened 2022-01-10
have minimal tests up-to-date and working

bjornaer:


pipeline/objects/decorators.py
```diff
@@ -34,7 +34,9 @@ def execute_func(*args, **kwargs):
                         )
                     )
 
-            node_output = Variable(type_class=function.__annotations__["return"])
+            node_output = Variable(
+                type_class=function.__annotations__["return"], is_output=True
```
paulcjh, line 38: I do not think this is correct, the is_output field is for when the variable is an output for the whole pipeline not a GraphNode. What's the motivation behind this?

paulcjh (changes requested):

pipeline/objects/decorators.py
bjornaer: One of the errors the test caught was `_output` [not being found within the variables set](https://github.com/mystic-ai/pipeline/blob/main/pipeline/objects/pipeline.py#L39) and this was raising an error, I thought perhaps when setting the functions we wanted to define as output their output

But from your comment I gather `is_output` is just for the whole pipeline not fro each node. I already have a change to fix the error while respecting this convention, just confirm if I understood correctly pls!

paulcjh: Okay gotcha, I thought I fixed this before actually my bad!
You can register it as an output node when builder.output(...) is called, this should be working it's weird that it's not.

bjornaer: yep, I'll review that bit to make it happen

paulcjh (approved):

PR #22, opened 2022-01-11
Max/add model schema

bjornaer:


paulcjh:
Model schemas look good.
Only issue I see here is with adding models as "Runnables". Only pipelines & functions should be "runnable" and then the model is referenced from either. It is not "Run" itself. Check out this HF example I made: https://github.com/mystic-ai/pipeline/blob/main/examples/hf_model.py

paulcjh (approved):
Great!

paulcjh:
I saw your comment on the FunctionGet, I think this is something Alex left. Will circle back, I think it's okay to delete but will double check.

alexpearce:
> I think this is something Alex left. Will circle back, I think it's okay to delete but will double check.

I think so too. The schemas in `function` are sufficient.

PR #27, opened 2022-01-17
Paul/auto docker

paulcjh:


examples/docker/basic_usage.py
```diff
@@ -0,0 +1,19 @@
+from pipeline import docker
+from pipeline import Pipeline, Variable, pipeline_function
+
+
+@pipeline_function
+def add_numbers(input_dict: dict) -> float:
+    return input_dict["in_1"] + input_dict["in_2"]
```
alexpearce, line 7: Is dictionary input a requirement for using a `Pipeline` with `create_pipeline_api`, or is this just an example and one could use `add_number(x: int, y:int)` instead?

pipeline/docker/__init__.py
```diff
@@ -0,0 +1,81 @@
...
+            "pipeline_redis": {
+                "image": "bitnami/redis:latest",
+                "container_name": "pipeline_redis",
+                "restart": "always",
+                "environment": ["ALLOW_EMPTY_PASSWORD=yes"],
+                "ports": ["6379:6379"],
+            },
+        },
+    }
+    with open(join(path, "docker-compose.yml"), "w") as docker_compose_file:
```
alexpearce, line 80: Why not use `os.path.join` here and remove the `ntpath` import? `ntpath` is Windows-specific.

pipeline/docker/__init__.py
```diff
@@ -0,0 +1,81 @@
...
+        paths.append(graph_path)
+
+    create_dockerfile(paths, output_dir=output_dir)
+    create_docker_compose(output_dir, **environment_variables)
+
+
+def create_dockerfile(pipeline_graph_paths: List[str], *, output_dir="./"):
+    with open(os.path.join(output_dir, "Dockerfile"), "w") as docker_file:
+        docker_file.writelines(
+            ["FROM --platform=linux/amd64 mysticai/pipeline-docker:latest"]
```
alexpearce, line 27: Is `--platform=linux/amd64` necessary? Can we not provide native containers? 😢

pipeline/docker/__init__.py
```diff
@@ -0,0 +1,81 @@
...
+        "DATABASE_CREATE_TABLES": 1,
+        "REDIS_URL": "pipeline_redis",
+    }
+
+    for key in presets_env_vars:
+        if key not in environment_vars:
+            environment_vars[key] = presets_env_vars[key]
+
+    docker_yml_dict = {
+        "version": "3.3",
```
alexpearce, line 53: Any particular reason to choose 3.3? If not we could go for the latest (3.9).

tests/test_docker.py
```diff
@@ -0,0 +1,32 @@
+import os
+
+from pipeline import docker
+from pipeline import Pipeline, Variable, pipeline_function
+
+
+def test_dockerfiles():
```
alexpearce, line 7: Nice!

examples/docker/hf_model.py
```diff
@@ -0,0 +1,42 @@
...
+    builder.add_variables(in_1)
+
+    gpt_neo = PipelineGPTNeo()
+
+    out_str = gpt_neo.predict(in_1)
+
+    builder.output(out_str)
+
+gpt_neo_pipeline = Pipeline.get_pipeline("GPTNeo")
+print(gpt_neo_pipeline.run("lol how are you mi amigo"))
```
bjornaer, line 41: you can do `Pipeline.run_local("GPTNeo", *args)` as well

alexpearce (commented):
Thanks! Looks like a great start. Some general questions from me.

pyproject.toml
```diff
@@ -55,3 +57,6 @@ exclude = '''
 
 [tool.isort]
 profile = "black"
+
+[tool.poetry.extras]
+docker = ["PyYAML", "docker"]
```
bjornaer, line 62: missing new line end of file

examples/docker/hf_model.py
alexpearce: Is there a need to have to 'fetch' the pipeline object? Could one also do `builder.run(…)`?

pipeline/docker/__init__.py
```diff
@@ -0,0 +1,81 @@
+from ntpath import join
+import os
+import yaml
+
+from typing import List
+
+from pipeline.objects import Graph
+
+
+def create_pipeline_api(
```
bjornaer, line 10: perhaps renaming this function? `dockerize_pipeline`?

also this could be moved directly to a `pipeline/docker.py`  

or this file and now that I think of it, the `decorators.py` could both be moved to `utils`

examples/docker/basic_usage.py
```diff
@@ -0,0 +1,19 @@
...
+with Pipeline("AddNumbers") as builder:
+    in_1 = Variable(dict, is_input=True)
+    builder.add_variables(in_1)
+
+    sum = add_numbers(in_1)
+
+    builder.output(sum)
+
+
+docker.create_pipeline_api([Pipeline.get_pipeline("AddNumbers")])
```
bjornaer, line 19: could we perhaps add a method under `Pipeline` abstracting this to be something like `Pipeline.dockerize(id: str)`

where it can invoke the functions from the `docker` module

pipeline/docker/__init__.py
bjornaer: +1

pipeline/docker/__init__.py
```diff
@@ -0,0 +1,81 @@
...
+        "services": {
+            "pipeline_api": {
+                "image": "mysticai/pipeline-docker:latest",
+                "container_name": "pipeline_api",
+                "restart": "always",
+                "ports": ["5010:80"],
+                "build": {"context": "./", "dockerfile": "Dockerfile"},
+                "environment": {
+                    **environment_vars,
+                },
```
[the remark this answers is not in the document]
bjornaer: +1 for the depends_on

examples/docker/hf_model.py
bjornaer: from builder you can also do `builder.run_local(id, *args)`

alexpearce: Seems nice. Is there a reason to re-fetch when that's possible then?

paulcjh: It's just a preference, and specifically here because I pass the pipeline as an arg to the docker building function the line after.
Personally I don't like the `.run_local` as it implies that there should be a `.run_remote` or an equivalent. The base pipeline features should always be locally focused.

examples/docker/basic_usage.py
paulcjh: I have actually changed the way args are sent now, so it was an issue with the inputs before. The full correction of this requires a change in the RunCreate schema to support generic inputs not just a str, which is what's now passed into the API. See [here](https://github.com/mystic-ai/pipeline-docker/blob/main/app/routes/run.py).

paulcjh: I will correct this in this example to pass in ints instead, but it will not work with the current pipeline package.

pipeline/docker/__init__.py
paulcjh: I missed this, you're very correct. I think I copied code from somewhere else and forgot to switch the import.

pipeline/docker/__init__.py
paulcjh: It's mostly for consistency, building on M1 will not work in some occasions or it can't then be passed on.
We can add this as a kwarg to the dockerize function so that people can select what OS they want.

pipeline/docker/__init__.py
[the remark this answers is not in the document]
paulcjh: Yup agreed.

pipeline/docker/__init__.py
```diff
@@ -0,0 +1,81 @@
...
+                "container_name": "pipeline_db",
+                "restart": "always",
+                "environment": {"POSTGRES_PASSWORD": "example"},
+            },
+            "pipeline_redis": {
+                "image": "bitnami/redis:latest",
+                "container_name": "pipeline_redis",
+                "restart": "always",
+                "environment": ["ALLOW_EMPTY_PASSWORD=yes"],
+                "ports": ["6379:6379"],
```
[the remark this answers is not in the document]
paulcjh: I think it's fine, will remove the Allow empty password though and go with a randomly generated one.

pyproject.toml
```diff
@@ -24,6 +24,8 @@ humps = "^0.2.2"
 pyhumps = "^3.0.2"
 requests-toolbelt = "^0.9.1"
 tqdm = "^4.62.3"
+PyYAML = {version = "^6.0", optional = true}
+docker = {version = "^5.0.3", optional = true}
```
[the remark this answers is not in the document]
paulcjh: Yes, I didn't end up using it.

tests/test_docker.py
paulcjh: lol

pipeline/docker/__init__.py
paulcjh: Naming is just a preference here, but happy to move into `docker.py`.

examples/docker/basic_usage.py
paulcjh: Happy to add another function for it, but still would like to keep this.
Personally i'm not a fan of doing it through `Pipeline` especially later when we start pulling pipelines from the cloud and then will have to probably use them as objects as well.

examples/docker/hf_model.py
bjornaer: yeah, I added a `Pipeline.run` which is remote, I think I misunderstood this as we wanted users to run on our cloud instead of locally by default

examples/docker/basic_usage.py
bjornaer: That is what I meant, add just an abstraction of this onto the pipeline, a sugarcoat.

when we pull them from the cloud, wont we pull dags(the graphs)? I had the understanding that `Pipeline` is a manager os all of the dags

pipeline/docker/__init__.py
bjornaer: Is it an api? I see we create a container, but are we are adding endpoints to send inputs and such? I ask because I see we pull an image for `pipeline-docker` which I don't know

examples/docker/basic_usage.py
bjornaer: it could be even written as `Pipeline.dockerize(ids: List[str])` and check for the graphs under the hood

pipeline/docker/__init__.py
paulcjh: Yes it is an API, link [here to the source](https://github.com/mystic-ai/pipeline-docker/blob/main/app/routes/run.py)

examples/docker/basic_usage.py
paulcjh: Gotcha, I must have missed that, we'll need to revert that my bad for not being clear

pipeline/docker/__init__.py
bjornaer: alright, now i understand, thanks!

examples/docker/basic_usage.py
bjornaer: no problem, we can also rename them, make local the default

examples/docker/hf_model.py
bjornaer: also if Pipeline managed the calls, you wouldn't need the extra fetch:
```
Pipeline.run(id: str, *args)
Pipeline.create_api(ids:List[str])
```

bjornaer: otherwise, for creating an api with many Graphs, you would need multiple fetches, would become quite verborragic, I think

examples/docker/basic_usage.py
bjornaer: > Happy to add another function for it, but still would like to keep this. Personally i'm not a fan of doing it through `Pipeline` especially later when we start pulling pipelines from the cloud and then will have to probably use them as objects as well.

this applies to the graphs -- which in the schemas are modeled under the name pipeline, but on our library logic the object called Pipeline is a `bag of dags` a high level manager of pipelines actually -- perhaps we should consider naming convention, but with the code as is, I think the best way to keep things clean is to manage pipelines via this high level manager object called `Pipeline`

alexpearce: I was pretty confused by the `Pipeline`/`Graph` difference, as to me 'Pipeline' is the graph whereas in this package the name for that is `Graph`. I would be in favour of dropping `Graph` and having everything under `Pipeline`, as things like `Pipeline.retrieve_remote` can be thought of as a 'constructor' of a `Pipeline`.

paulcjh:
@bjornaer from your two comment threads I see that you want to include more thing in the Pipeline top level class and direct the user to use most low level features through there. Personally I'm not a fan of this and I think long term it will bulk up the Class, if not in the short-term. I do have a preference to keep things separate: if you want to create a docker API then you go to a docker object, if you want to talk to PipelineCloud then you talk to the PipelineCloud object.

So from here I think we should setup a meeting (with @alexpearce) to discuss how much we want to include in the Pipeline class; but for the moment I think we should merge this PR and then add in the adjustments later based on the outcomes of that meeting, how does this sound? I'm also suggesting this as your comments are for additions rather than adjustments on the changes I've added.

bjornaer:
> @bjornaer from your two comment threads I see that you want to include more thing in the Pipeline top level class and direct the user to use most low level features through there. Personally I'm not a fan of this and I think long term it will bulk up the Class, if not in the short-term. I do have a preference to keep things separate: if you want to create a docker API then you go to a docker object, if you want to talk to PipelineCloud then you talk to the PipelineCloud object.
> 
> So from here I think we should setup a meeting (with @alexpearce) to discuss how much we want to include in the Pipeline class; but for the moment I think we should merge this PR and then add in the adjustments later based on the outcomes of that meeting, how does this sound? I'm also suggesting this as your comments are for additions rather than adjustments on the changes I've added.

alright, I am resolving the convos, please setup a meeting for us to talk about it!

bjornaer (approved):

PR #29, opened 2022-01-20
Correct Top URL

alexpearce:
The trailing slash results in a redirect response, but the redirect doesn't include the `Authorization` header in production so causes 403 errors.

(Most Top URLs do not end in a slash; `POST /v2/files/` is one exception I know of, there may be others)


bjornaer (dismissed):

bjornaer (approved):
please update the code to pass linting 
LGTM

bjornaer (approved):

PR #32, opened 2022-01-28
Optional camelcase aliases

plutopulp:
Added option for `CAMEL_CASE_ALIASES` being set from environment.


pipeline/schemas/base.py
```diff
@@ -10,7 +11,7 @@
 #: When set to True, model fields can be get/set by camelCase'd field names
 #: This is useful for JS interop., where camelCase is the convention
 #: (rather than Python's snake_case convention)
-CAMEL_CASE_ALIASES = False
+CAMEL_CASE_ALIASES = os.environ.get("CAMEL_CASE_ALIASES")
```
alexpearce, line 14: Let's have a more explicit boolean here.

```suggestion
CAMEL_CASE_ALIASES = bool(os.environ.get("CAMEL_CASE_ALIASES", False))
```

(Or you could covert this to a single-field Pydantic `BaseSettings` object.)

alexpearce (changes requested):

pipeline/schemas/base.py
plutopulp: Cool, thx man! With just the single env variable, leaving it with your suggestion for now.

alexpearce (approved):
Thanks pal!

PR #41, opened 2022-02-11
Add schemas for runnable metrics endpoints

andrei-trandafir:


alexpearce (dismissed):
LGTM! Thanks for this. A bit of documentation on each schema would be nice but I won't insist.

paulcjh (approved):

alexpearce (approved):
Thanks for the docs! Looks good to me.

Only thing we could improve on is mentioning that some metrics, e.g. averages, are undefined when `run_count` is zero.

PR #42, opened 2022-02-11
Fix

paulcjh:


pipeline/objects/graph.py
```diff
@@ -165,7 +165,7 @@ def from_schema(cls, schema: PipelineGet):
                         "Did not find a class to bind for model (local_id:%s)"
                         % model.local_id
                     )
-            else:
+            elif _func.class_instance != None:
```
alexpearce, line 168:
```suggestion
            elif _func.class_instance is not None:
```

alexpearce (changes requested):

pipeline/objects/graph.py
paulcjh: `class_instance` is used when a function has a `self` arg (it's a part of a class). So `class_instance` is None when the function is not in the context of a class.

alexpearce (approved):

PR #43, opened 2022-02-11
docker module test change

bjornaer:
make test autoclean created files

enable arbitrary directory to be provided for docker api functionality

fixed pattern in codeowners file

renamed `run_local` to just `run`

### DIFF
--- a/pipeline/docker/__init__.py
+++ b/pipeline/docker/__init__.py
@@ -1,7 +1,8 @@
 import os
-import yaml
-
+from pathlib import Path
 from typing import List
+
+import yaml
 
 from pipeline.objects import Graph
 
@@ -14,9 +15,10 @@ def create_pipeline_api(
     **environment_variables,
 ):
     paths = []
+    Path(output_dir).mkdir(parents=True, exist_ok=True)
 
     for pipeline_graph in pipeline_graphs:
-        graph_path = pipeline_graph.name + ".graph"
+        graph_path = os.path.join(output_dir, pipeline_graph.name + ".graph")
         pipeline_graph.save(graph_path)
         paths.append(graph_path)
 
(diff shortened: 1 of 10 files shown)


pipeline/docker/__init__.py
```diff
@@ -14,9 +15,10 @@ def create_pipeline_api(
     **environment_variables,
 ):
     paths = []
+    Path(output_dir).mkdir(parents=True, exist_ok=True)
 
     for pipeline_graph in pipeline_graphs:
-        graph_path = pipeline_graph.name + ".graph"
+        graph_path = os.path.join(output_dir, pipeline_graph.name + ".graph")
```
alexpearce, line 21: nbd but I always thought `pathlib.Path`'s overload of the `/` operator was kinda cute 😄 So we could do

```python
p = Path(output_dir)
p.mkdir(parents=True, exist_ok=True)
# …
graph_path = output_dir / pipeline_graph.name + ".graph"
```

tests/test_docker.py
```diff
@@ -22,11 +22,17 @@ def add_numbers(in_1: float, in_2: float) -> float:
...
-    assert os.path.exists("./Dockerfile") and os.path.exists("./docker-compose.yml")
+    assert os.path.exists(f"{test_dir}/Dockerfile") and os.path.exists(
+        f"{test_dir}/docker-compose.yml"
+    )
     # Check for the serialization of the graph
-    assert os.path.exists("./AddNumbers.graph")
+    assert os.path.exists(f"{test_dir}/AddNumbers.graph")
+
+    if os.path.exists(test_dir):
+        shutil.rmtree(test_dir)
```
alexpearce, line 36: Also nbd but you could leverage [pytest's existing tmpdir support](https://docs.pytest.org/en/6.2.x/tmpdir.html) to avoid having to do the cleanup by hand.

alexpearce (approved):
LGTM! Only a couple of suggestions from me but am happy for you to ignore them.

pipeline/docker/__init__.py
bjornaer: I feel it can be obscure to the unaware dev, if you have strong feelings about it I can change though

tests/test_docker.py
bjornaer: aye aye hahah

pipeline/docker/__init__.py
alexpearce: Nah not strong.

alexpearce (approved):

PR #44, opened 2022-02-14
Paul/multi input runs

paulcjh:


paulcjh:
Addition of GPT J showing multi inputs
Multi inputs
Non str input types for runs

pipeline/api/cloud.py
```diff
@@ -57,6 +58,11 @@ def upload_file(self, file_or_path, remote_path) -> FileGet:
         else:
             return self._post_file("/v2/files/", file_or_path, remote_path)
 
+    def upload_data(self, file_or_path, remote_path) -> DataGet:
+        uploaded_file = self.upload_file(file_or_path, remote_path)
+        uploaded_data = self._post("/v2/data/", uploaded_file.dict())
```
alexpearce, line 63: Could you remove the trailing slash here? Most top routes do not have one (the exception being `POST /v2/files/`, which we should fix).

alexpearce (approved):
Just a URL change please.

pipeline/api/cloud.py
paulcjh: Sure, and will change on the top node too (I added a "/" to match files)

alexpearce (approved):

PR #47, opened 2022-02-17
Require `total` parameter when constructing Paginated

alexpearce:
Currently `Paginated.of` computes `Paginated.total` by counting the number of items it receives, but `Paginated.total` should represent the total number of objects that are available for pagination, rather than the number currently being returned.

This PR modifies `Paginated.of` to _require_ a `total: int` parameter, which is used to populate the `Paginated.total` field.

Towards [NA-389](https://neuro-ai.atlassian.net/browse/NA-389).


andrei-trandafir (approved):
Good spot!

PR #53, opened 2022-02-23
Additional token patch fields

paulcjh:


isNeil (approved):
LGTM

pipeline/schemas/token.py
```diff
@@ -40,6 +40,10 @@ class TokenPatch(Patchable):
     name: Optional[str]
     # Boolean specifying if token can be used
     is_enabled: Optional[bool]
+    # Token value (only sudo tokens can use this)
+    token: Optional[str]
+    # Datetime to expire the token
+    expire_at: Optional[datetime]
```
alexpearce, line 46: @plutopulp and I had a chat about this and concluded that letting this field be patched defeats the purpose of having it. If a token has an expiry date but gets leaked the attacker can just set the expiry date to far in the future using the leaked token.

GitHub provides a feature to create a new token with an identical feature set but with a new expiry and value. Going down that route feels nicer to me.

pipeline/schemas/token.py
```diff
@@ -40,6 +40,10 @@ class TokenPatch(Patchable):
     name: Optional[str]
     # Boolean specifying if token can be used
     is_enabled: Optional[bool]
+    # Token value (only sudo tokens can use this)
+    token: Optional[str]
```
alexpearce, line 44: Do you have a corresponding pipeline-top PR for this? What's the use case here?

paulcjh: APIv1 migration

paulcjh: I need to be able to set user's tokens to what they were in APIv1

pipeline/schemas/token.py
paulcjh: Do all tokens have the same permissions for users?
A reasonable use case here would be if someone needs to extend a production token expiry date.

paulcjh: Also it's possible that the expire_at field can only be set by dashboard token

PR #55, opened 2022-02-23
Add organisation member invite schemas

alexpearce:
Towards [NA-419](https://neuro-ai.atlassian.net/browse/NA-419).

### DIFF
--- a/pipeline/schemas/organisation_member_invite.py
+++ b/pipeline/schemas/organisation_member_invite.py
@@ -1,0 +1,25 @@
+from .base import BaseModel
+
+
+class OrganisationMemberInviteBase(BaseModel):
+    #: The ID of the Organisation the invite relates to
+    organisation_id: str
+    #: The email the invite is to be sent to
+    email: str
+
+
+class OrganisationMemberInviteCreate(OrganisationMemberInviteBase):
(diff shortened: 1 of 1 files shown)


pipeline/schemas/organisation_member_invite.py
```diff
@@ -0,0 +1,23 @@
...
+    pass
+
+
+class OrganisationMemberInviteGet(OrganisationMemberInviteBase):
+    """View of an invitation to join an Organisation."""
+
+    #: The ID of this invite
+    id: str
+    #: If True this invite has been accepted (emailed link has been visited)
+    accepted: bool
```
plutopulp, line 23: The dashboard conditionally sends a email template depending on whether user is registered or not. Would be good to also include this field

alexpearce: Great spot! Added a `has_user` field.

plutopulp: Maybe we can change this name in the future, as I find it a bit misleading given that the field is indicating whether a user with the invitee email exists in the User table. But I don't feel too strongly about it.

plutopulp (approved):

PR #57, opened 2022-02-24
Remove humps dependency

plutopulp:
The [humps package](https://pypi.org/project/humps/) provides the same module as the [pyhumps package](https://pypi.org/project/pyhumps/). We want the latter but the dependencies specify both for some reason.

This PR removes humps, which removes possible ambiguities when importing `humps`.


alexpearce:
Just a heads-up that `poetry lock` will try to update all packages within the constraints specified in the `pyproject.toml`, which is why there are unrelated changes in the lockfile. You can use `poetry lock --no-update` to avoid this (although it's typically no big deal).

alexpearce (approved):
Thanks!

PR #65, opened 2022-03-16
Update validators.py

plutopulp:
This allows for a more extensive set of passwords

### DIFF
--- a/pipeline/schemas/validators.py
+++ b/pipeline/schemas/validators.py
@@ -15,7 +15,7 @@ EMAIL_REGEX = re.compile(
 )
 # Simple password regex, requires a minimum of 8 characters, one uppercase letter,
 # one lowercase letter and a number.
-PASSWORD_REGEX = re.compile(r"^(?=.*[a-z])(?=.*[A-Z])(?=.*\d)[a-zA-Z\d]{8,}$")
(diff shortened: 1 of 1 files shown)


pipeline/schemas/validators.py
```diff
@@ -15,7 +15,7 @@
 )
 # Simple password regex, requires a minimum of 8 characters, one uppercase letter,
 # one lowercase letter and a number.
```
alexpearce, line 17: This comment should be updated to reflect the new rules.

alexpearce:
```suggestion
# Simple password regex, requires a minimum of 8 characters with at least one
# uppercase letter, one lowercase letter, and one number.
```

paulcjh (approved):

tests/schemas/test_validators.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def test_invalid_password():
+    for password in INVALID_PASSWORDS:
+        assert valid_password(password) == False
+
+
+def test_valid_password():
+    for password in VALID_PASSWORDS:
+        assert valid_password(password) == True
```
alexpearce, line 25: Obviously NBD but convention for asserting booleans would be plain `assert valid_password(password)` and `assert not valid_password(password)` ☄️

PR #70, opened 2022-03-29
Projects

paulcjh:
Currently every interaction with the API puts all data (where appropriate) into the Default project.
This PR allows users to create, and use their own projects from the pipeline library.

### DIFF
--- a/pipeline/api/cloud.py
+++ b/pipeline/api/cloud.py
@@ -16,6 +16,7 @@ from pipeline.schemas.function import FunctionCreate, FunctionGet
 from pipeline.schemas.model import ModelCreate, ModelGet
 from pipeline.schemas.pipeline import PipelineCreate, PipelineGet, PipelineVariableGet
 from pipeline.schemas.run import RunCreate
+from pipeline.schemas.project import ProjectGet, ProjectCreate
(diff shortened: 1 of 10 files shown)


pipeline/api/cloud.py
```diff
@@ -51,6 +53,12 @@ def authenticate(self, token: str = None):
             print("Succesfully authenticated with the Pipeline API (%s)" % self.url)
         self.token = _token
 
+        if self.token and project:
+            try:
+                self.active_project = self.create_project(project)
+            except:
```
bjornaer, line 59: this is a bit weird, if create project fails, what error is it generating? in the `create_project` function code the only place that seems to be able to raise an exception is the `parse_obj` function -- the post might return 400's but they will come back as json responses -- then you want to try a get to get the project and set it as active?

what if the same exception comes from `parse_obj`?

also, if this is because if project exists POST might fail, you could just get first, if exists you done, otherwise POST to create, in a `create_if_not_exists` sort of pattern

Just trying to understand the logic behind this `try-except`  block

bjornaer (changes requested):
one question, otherwise seems alright...

you do have failing tests and conflict with main though!

alexpearce:
Closing this while we focus on single project workflows. Can revisit when we return to having multiple projects.

PR #72, opened 2022-03-31
add a few tests for pipeline cloud

bjornaer:
and as a side effect of some commits without linting/format checks I had to cleanup some flake8 errors


alexpearce (approved):
Always nice to have a few more tests! Thanks a lot for this, and for the lining fix-ups. I guess we're not running those in the CI then?

PR #73, opened 2022-03-31
fixing small details on dev setup script

bjornaer:


andrei-trandafir (dismissed):
LGTM

setup.sh
```diff
@@ -12,11 +14,10 @@ chmod +x .git/hooks/pre-push
 echo "${G}Pre-Push hooks set.${NC}"
 
 echo "${Y}Installing deps...${NC}"
-poetry shell
-poetry install
+poetry run install
```
alexpearce, line 17: I think this one should stay as `poetry install`, otherwise `poetry run` will try to execute the `install` program (typically at `/usr/bin/install`).

bjornaer: I did run it and it seemed alright 🤔 maybe cause I had an already set env?

alexpearce: I meant that it will run but won't be doing what you expect.

```
$ poetry run install
usage: install [-bCcpSsv] [-B suffix] [-f flags] [-g group] [-m mode]
               [-o owner] file1 file2
       install [-bCcpSsv] [-B suffix] [-f flags] [-g group] [-m mode]
               [-o owner] file1 ... fileN directory
       install -d [-v] [-g group] [-m mode] [-o owner] directory ...
```

I think `poetry install` fits the intention ("Installing deps...").

alexpearce (changes requested):
A few fix-ups:

- `chmod +x setup.sh`
- Change `poetry run install` to `poetry install`
- Add `set -euo pipefail` to enable ['bash strict mode'](http://redsymbol.net/articles/unofficial-bash-strict-mode/), which will stop script execution if any one command fails.

README.md
```diff
@@ -99,7 +99,7 @@ python -m pip install -U pipeline-ai
 This project is made with poetry, [so firstly setup poetry on your machine](https://python-poetry.org/docs/#installation).
 
 Once that is done, please run
-
+    chmod +x setup.sh
```
alexpearce, line 102: We can do this to the actual file in the repository and commit it, then users won't have to do it themselves.

alexpearce (changes requested):

README.md
bjornaer: did not know git would track this

README.md
```diff
@@ -99,7 +99,6 @@ python -m pip install -U pipeline-ai
 This project is made with poetry, [so firstly setup poetry on your machine](https://python-poetry.org/docs/#installation).
 
 Once that is done, please run
-
```
alexpearce, line 102: Removing this messes up the code block formatting.

alexpearce (approved):
Nice one! Thanks.

PR #75, opened 2022-04-07
Add schemas for hardware related metrics

andrei-trandafir:
Towards [NA-480](https://neuro-ai.atlassian.net/browse/NA-480).


alexpearce (approved):
All seems reasonable to me. Thanks!

PR #80, opened 2022-04-20
[NA-497] - add new model overview schema

bjornaer:
adding new schema with specific details for model hub dashboard

### DIFF
--- a/pipeline/schemas/model.py
+++ b/pipeline/schemas/model.py
@@ -12,7 +12,6 @@ class ModelBase(BaseModel):
 
 
 class ModelGet(ModelBase):
-    id: str
     hex_file: FileGet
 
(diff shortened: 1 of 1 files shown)


pipeline/schemas/model.py
```diff
@@ -25,6 +24,11 @@ class ModelGetDetailed(ModelGet):
     ...
 
 
+class ModelGetOverview(ModelBase):
+    description: str = ""
+    pipeline_count: int = 0
```
alexpearce, line 29: I'd be tempted to remove the default. This value is a big part of why this schema exists, so I'd like the creator of the instance to be forced to fill it in themselves.

alexpearce (approved):
LGTM! Thanks very much.

PR #84, opened 2022-04-23
Deployment schemas (NA-513)

paulcjh:
The changes in this PR are relevant to NA-513, NA-521/522/523/524.
There are 3 schema changes, specifically: `DeploymentGet`, `DeploymentCreate` & `DeploymentPatch`.

### DIFF
--- a/pipeline/schemas/pipeline.py
+++ b/pipeline/schemas/pipeline.py
@@ -72,7 +72,6 @@ class PipelineCreate(BaseModel):
 
 
 class PipelineDeployCreate(BaseModel):
-    pipeline_id: str
     project_id: str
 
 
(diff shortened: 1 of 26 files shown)


pipeline/schemas/pipeline.py
```diff
@@ -56,6 +57,7 @@ class PipelineGet(PipelineGetBrief, RunnableGet):
     models: List[ModelGet]
     graph_nodes: List[PipelineGraphNode]
     outputs: List[str]
+    project: ProjectGet
```
alexpearce, line 60: I think we can remove this for now so we don't have to deal with permissions for nested objects which may belong to different users.

paulcjh: done

alexpearce (approved):

PR #88, opened 2022-05-11
Extend PipelineCreate fields

alexpearce:
1. Adds some of the fields recently added to Pipeline's API to the Pipeline creation schema.
2. Allows setting the new fields when uploading a Pipeline.
3. Add a `run_examples` field to detailed schema to support language-dependent usage snippets.
4. Bump version.

### DIFF
--- a/pipeline/api/cloud.py
+++ b/pipeline/api/cloud.py
@@ -4,7 +4,7 @@ import io
 import json
 import os
 import urllib.parse
-from typing import TYPE_CHECKING, Any, List, Optional, Union
+from typing import TYPE_CHECKING, Any, List, Optional, Set, Union
 
 import requests
 from requests_toolbelt.multipart import encoder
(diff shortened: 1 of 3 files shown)


andrei-trandafir (dismissed):

pipeline/api/cloud.py
```diff
@@ -212,6 +218,9 @@ def upload_pipeline(self, new_pipeline_graph: Graph) -> PipelineGet:
             models=new_models,
             graph_nodes=new_graph_nodes,
             outputs=new_outputs,
+            public=public,
+            description=description,
+            tags=set() if tags is None else tags,
```
andrei-trandafir, line 223: Possible change this to `tags or set()`? May seem simpler, your choice.

andrei-trandafir (approved):

pipeline/api/cloud.py
alexpearce: 💯

andrei-trandafir (approved):
🛳️

PR #109, opened 2022-06-15
[BE-59] - download function and unpickle hex

bjornaer:
Enable retrieving `Get` schemas from Pipeline-Top endpoints.

Once the data has been retrieved via the function get endpoint we must reform the python object from the hex string in the FileGet schema.


isNeil (approved):
LGTM!

PR #110, opened 2022-06-16
[BE-61][BE-62]Enable Model and Data download

bjornaer:
adding functionality to download models and data plus some small tweaks to how functions were previously downloaded

plus unit tests


isNeil (approved):
LGTM nice refactor

PR #113, opened 2022-07-04
[BE-104] Fix pipeline examples

rossgray:
Currently, quite a few of the examples are failing. This PR fixes these. I'm not too familiar with all the expected behaviour of this library so any help here would be greatly appreciated.


examples/api_pipeline_run.py
```diff
@@ -22,8 +20,6 @@ def add_lol(a: str) -> str:
 
 run_result = api.run_pipeline(upload_output, "Hi I like to")
 print("Run state:", run_result["run_state"])
-try:
-    result_preview = json.loads(run_result["result_preview"])
-except (TypeError, json.JSONDecodeError):
-    result_preview = "unavailable"
+
+result_preview = run_result["result_preview"]
```
rossgray, line 24: result preview is a list so JSON decoding was failing

alexpearce: It might be that the `try`/`except` handles the case of a failed run. The Run API doesn't offer a very clear contract in this case at the moment.

examples/basic_maths/README.md
```diff
@@ -74,5 +79,4 @@ with Pipeline("MathsIsFun") as pipeline:
 
 output_pipeline = Pipeline.get_pipeline("MathsIsFun")
 print(output_pipeline.run(5.0, 6.0))
-
 ```
```
alexpearce, line 82: Well done for remembering to check the README 😆 so easy to miss.

examples/huggingface/gpt-j-6b.py
```diff
@@ -130,6 +130,7 @@ def __exit__(self, type, value, traceback):
                 self.tokenizer = AutoTokenizer.from_pretrained(self.tokenizer_path)
 
 
+# NOTE: requires PIPELINE_API_TOKEN environment variable to be set
 api = PipelineCloud()
```
alexpearce, line 134: Alternatively we could pass in an obvious placeholder token (no strong preference from me).

```suggestion
# Replace this token with your own
api = PipelineCloud(token="pipeline_token_value")
```

alexpearce (commented):
Thanks very much for tidying things up!

I guess some (all?) of these require access to the API, otherwise it would be nice to automate the testing of these to avoid them drifting away from the spec in the future.

examples/api_pipeline_run.py
rossgray: Yeah I'm not sure to be honest. Looking at the [type annotation](https://github.com/mystic-ai/pipeline/blob/main/pipeline/schemas/run.py#L92) for the `result_preview` it is supposed to be a `list` or `dict` and `json.loads` expects a `str` so this would never work. Perhaps we were trying to serialise the `dict` into a string instead? In any case, I will ignore the JSON stuff and add a `try/except` to catch the case where there is no `result_preview`

examples/huggingface/gpt-j-6b.py
rossgray: Thanks. I've added your suggestion but also noted that they can alternatively use the env var

rossgray:
Thanks @alexpearce for reviewing. I've done a bit more cleanup:
- I've deleted the old examples that we don't want to support currently
- I've combined `examples/api_usage.py` and `examples/api_pipeline_run.py` and deleted the latter, since there was a lot of overlap between the two and I think just having one clear example is better
- I've applied some changes based on your feedback (API tokens + `try/except` handling)

> I guess some (all?) of these require access to the API, otherwise it would be nice to automate the testing of these to avoid them drifting away from the spec in the future.
Most, but not all, required API access. We could still add some tests by mocking parts of the API client. Let me know if you'd like me to do this or not.

alexpearce (approved):
👏

rossgray:
sorry @alexpearce, I needed to bump the version in `pyproject.toml`

alexpearce (approved):

PR #116, opened 2022-07-12
[BE-103][BE-105] Brendan/handle undeployed pipeline

brendanbennett:
A fairly crude handling of the error created when a user tries to run a pipeline they haven't deployed. Specifically I've written a generic handler that spits out the message that's sent by Top when before it was just being handled as an HTTP error with no helpful message. The issue is that we have many errors that arise in Top that are sent as generic HTTP 4xx errors. A much better implementation would be to GET request Top first to check their pipeline is deployed, since this is potentially a very common user error. Then we can easily handle this specific case without resorting to a generic handler.

### DIFF
--- a/pipeline/api/cloud.py
+++ b/pipeline/api/cloud.py
@@ -361,9 +361,6 @@ class PipelineCloud:
             )
 
         run_create_schema = RunCreate(pipeline_id=pipeline_id, data_id=_data_id)
-        # TODO Create a GET request that checks the pipeline is deployed and
-        # handle this separately if not. Currently is handled crudely in a
-        # generic Exception in _post()
         return self._post("/v2/runs", json.loads(run_create_schema.json()))
 
     def _download_schema(
--- a/pipeline/api/cloud.py
+++ b/pipeline/api/cloud.py
@@ -137,8 +137,6 @@ class PipelineCloud:
             try:
                 response.raise_for_status()
             except:
-                # TODO Create a GET request that checks that a pipeline is deployed
-                # before running it.
                 content = json.loads(response.content.decode("utf-8"))
                 detail = content["detail"] if "detail" in content.keys() else UNKNOWN_ERROR_MESSAGE
                 raise Exception(f"Error {response.status_code}: {str(detail)}")
@@ -354,6 +352,9 @@ class PipelineCloud:
             )
 
         run_create_schema = RunCreate(pipeline_id=pipeline_id, data_id=_data_id)
+        # TODO Create a GET request that checks the pipeline is deployed and 
+        # handle this separately if not. Currently is handled crudely in a
+        # generic Exception in _post()
         return self._post("/v2/runs", json.loads(run_create_schema.json()))
 
     def _download_schema(
--- a/pipeline/api/cloud.py
+++ b/pipeline/api/cloud.py
@@ -36,6 +36,7 @@ UNKNOWN_ERROR_MESSAGE = (
     "There was an internal error. If this persists please contact support@mystic.ai"
 )
 
+
 class PipelineCloud:
     token: Optional[str]
     url: Optional[str]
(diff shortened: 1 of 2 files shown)


brendanbennett:
New functionality is that any HTTP error that is received from Top is first checked for a detail attribute. If one is found then it is a specific Error (i.e. that is accounted for on Top) and contains useful information. The Exception that is raised then contains this information. Otherwise it is an unexpected error (any network error, 404 etc.) and is treated as any other error and raises an HTTPError with only limited information. Previously, any extra information from Top about the error was lost when it was handled by the built in HTTPError handler.

pipeline/api/cloud.py
```diff
@@ -89,6 +93,19 @@ def raise_for_invalid_token(self):
...
+    @staticmethod
+    def _get_raise_for_status(response: requests.Response) -> None:
+        # A handler for errors that might be sent with messages from Top.
+        if not response.ok:
+            content = response.json()
+            if "detail" in content.keys() and isinstance(content["detail"], Dict):
+                detail = content["detail"]
+                message = detail.pop("message", "Unidentified Error")
+                detail = detail or ""
+                raise Exception(f"Error {response.status_code}: {message} {detail}")
```
alexpearce, line 105: The general behaviour here is nice, just a few comments on the details.

1. The idiomatic way of checking membership would be `"detail" in content`.
2. There are a lot of checks on the `detail` key.  I guess this is partly because Top does not have a consistent error message format? So sometimes it has `dict(detail="error message")` and sometimes it has `dict(detail=dict(message="error message"))`.

brendanbennett: Yes, as suspected some exceptions only send a string in detail. I will implement this. Also some exceptions only send a message with no other info

brendanbennett: PermissionNotGranted for example

alexpearce: OK, that's a bit annoying, and no doubt an oversight on our part not to design a consistent error interface.

Something I forgot: it would be good to raise a more specific error here. ATM the code is so generic it'd be annoying to specialise the exception class per error, but perhaps having something like `class PipelineAPIRequestError(Exception): pass` would be a start?

brendanbennett: I'm just finding out how gross the exceptions are in Top

pipeline/api/cloud.py
```diff
@@ -89,6 +89,30 @@ def raise_for_invalid_token(self):
...
+            detail = content.pop("detail", "")
+            message = None
+            # In most cases detail is not a string but a dict.
+            if isinstance(detail, dict):
+                message = detail.pop("message", None)
+            elif not isinstance(detail, str):
+                detail = ""
+            detail = detail or ""
+            # If a message was delivered we want to show that. Otherwise it's a
+            # standard HTTP error and should be handled my raise_for_status()
```
alexpearce, line 110:
```suggestion
            # standard HTTP error and should be handled by raise_for_status()
```

pipeline/api/cloud.py
```diff
@@ -89,6 +89,30 @@ def raise_for_invalid_token(self):
...
             )
 
+    @staticmethod
+    def _get_raise_for_status(response: requests.Response) -> None:
+        # A handler for errors that might be sent with messages from Top.
+        if not response.ok:
+            content = response.json()
+            # Every exception has content of {detail, status_code[, headers]}
+            # TODO Some exceptions in Top send detail as a string and not a dict.
+            # These exceptions are now handled like normal HTTP excpetions.
```
alexpearce, line 99:
```suggestion
            # These exceptions are now handled like normal HTTP exceptions.
```

alexpearce (approved):
This is a nice improvement! It'll look even nicer one the Top exception formats are consistent.

brendanbennett:
> This is a nice improvement! It'll look even nicer one the Top exception formats are consistent.

https://github.com/mystic-ai/pipeline-top/pull/228

alexpearce (approved):

pipeline/api/cloud.py
```diff
@@ -89,6 +89,30 @@ def raise_for_invalid_token(self):
...
+            # These exceptions are now handled like normal HTTP exceptions.
+            # Need to rewrite these to all have the same format.
+            detail = content.pop("detail", "")
+            message = None
+            # In most cases detail is not a string but a dict.
+            if isinstance(detail, dict):
+                message = detail.pop("message", None)
+            elif not isinstance(detail, str):
+                detail = ""
+            detail = detail or ""
```
isNeil, line 108: this line is redundant for future reference

PR #117, opened 2022-08-02
Add concept of compute requirements for a Run, Pipeline and Function

rossgray:
## Changes

- Added `ComputeRequirements` for `Run` (which can be given defaults at the `Pipeline`/`Function` level)
- Added `compute_type` to `Pipeline`s and `Function`s too (since it feels sensible if they can specify compute requirements they should also be able to specify the compute type)

## Questions/Comments

- So far I've only added a single optional compute requirement, `min_gpu_vram_mb` (not sure if this is the best name or if just `min_vram_mb` would be better?)
- Not sure whether to add these fields to the response schemas such as `PipelineGet`?

### DIFF
--- a/pipeline/schemas/run.py
+++ b/pipeline/schemas/run.py
@@ -4,7 +4,7 @@ from typing import Any, List, Optional, Union
 
 from pydantic import root_validator
 
-from pipeline.schemas.compute_requirements import ComputeRequirements
+from pipeline.schemas.compute_requirements import ComputeRequirements, ComputeType
 from pipeline.schemas.file import FileGet
 from pipeline.schemas.function import FunctionGet, FunctionGetDetailed
 from pipeline.schemas.pipeline import PipelineGet, PipelineGetDetailed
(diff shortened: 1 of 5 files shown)


pipeline/schemas/run.py
```diff
@@ -48,6 +43,7 @@ class RunCreate(BaseModel):
     blocking: Optional[bool] = False
     # By default a Run will require GPU resources
     compute_type: ComputeType = ComputeType.gpu
```
alexpearce, line 45: I think we're planning over having a Run's requirements 'override' those given in the runnable. The `compute_type` field should then also be optional, a missing value meaning 'using the requirements from the runnable'.

alexpearce (commented):
The schemas look reasonable to me, thanks!

Have you thought about the Run/Pipeline creation interface for specifying requirements? Do you want to work on that here or a separate PR?

pipeline/schemas/run.py
rossgray: Good point, have fixed that 👍

rossgray:
> Have you thought about the Run/Pipeline creation interface for specifying requirements? Do you want to work on that here or a separate PR?

Will probably do this in a follow up PR

alexpearce (approved):
:shipit:

PR #118, opened 2022-08-09
Allow clients to specify compute requirements

rossgray:
Allows users to specify the compute type (CPU/GPU) and additional GPU compute requirements at both a Pipeline and individual Run level.

### DIFF
--- a/pipeline/schemas/pipeline.py
+++ b/pipeline/schemas/pipeline.py
@@ -1,7 +1,7 @@
 from datetime import datetime
 from typing import Dict, List, Optional, Set
 
-from pydantic import Field, root_validator
+from pydantic import Field, root_validator, validator
 
 from pipeline.schemas.base import BaseModel
 from pipeline.schemas.compute_requirements import ComputeRequirements, ComputeType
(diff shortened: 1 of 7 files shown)


pipeline/schemas/pipeline.py
```diff
@@ -88,3 +88,15 @@ class PipelineCreate(BaseModel):
...
+    def compute_type_is_gpu(cls, v, values):
+        """If compute_type is not 'gpu' we don't expect any additional compute
+        requirements to be specified.
+        """
+        if values["compute_type"] != ComputeType.gpu:
+            if v and v.min_gpu_vram_mb:
+                raise ValueError(
+                    "min_gpu_vram_mb should only be specified for gpu workloads"
+                )
+        return v
```
alexpearce, line 102: Neat way of codifying the constraints. Hopefully we can design the OpenAPI spec in a way that the generated schemas contain such checks.

alexpearce (approved):
Looks great to me, thanks for the work.

How about a test that verifies the run-overrides-pipeline logic?

rossgray:
Thanks for the review @alexpearce

> How about a test that verifies the run-overrides-pipeline logic?
This logic is all contained within the backend when the run is submitted

PR #119, opened 2022-08-10
Add in PipelineFiles for loading in model weights

paulcjh:
At the moment if a model has a large weights file that it requires to run (see our DALLE models) we have to send a `requests` request to a server containing the weights, that we have manually setup, in the model's load function.

This PR implements the `PipelineFile` class which allows a user to to resolve the above issues.

How it works:
The `PipelineFile` class takes in a `path` kwarg which is the path to that file on the user's system. When running locally, the user can access this path to use arbitrarily. The `PipelineFile` object extends the `Variable` class and is treated as one in the computational graph.
When uploading the pipeline to PipelineCloud, the local file referred to in the `path` kwarg is uploaded as a File object on the API. When a Run request is put against the pipeline the `PipelineFile` object is slightly modified before the pipeline is run. The `path` variable in the `PipelineFile` object is modified to reflect the new path on the local system where the file has been loaded (the file object is retrieved and stored physically on a drive, or if small enough in RAM).

This PR also introduces the `torch_utils` file with some common functions that are used in PipelineLibrary when working with pytorch.

TODO:

- [x] Create locally working `PipelineFile` example
- [x] Complete uploading of the `PipelineFile` to PipelineCloud
- [x] Update the `PipelineVariableGet` schema to include optional fields for the `PipelineFile`
- [x] Complete reforming of the Graph on the resource
- [x] Add back compatibility for the `.load(...)` function in `PipelineModels`

NOTES:
1. This is a change to the cacheing mechanism. Previously the _load() function in the graph was called that then subsequently called each `PipelineModel`'s load function. This has been removed and the user is expected to add a `on_startup` field to an arbitrary function if they want it to be treated as the previous load function.


alexpearce:
Good stuff! I consider [Flyte](https://docs.flyte.org/en/latest/index.html) to have a pretty well thought-out API and this is similar in style to [`FlyteFile`](https://docs.flyte.org/projects/cookbook/en/latest/auto/core/flyte_basics/files.html), so :+1:

An aside to the client-side implementation, but related to the feature itself: if we're going to support uploading potentially large files now might be the right time to implement some form of hash-based, user-scoped de-duplication to avoid scripts uploading the same file contents again and again.

paulcjh:
> Good stuff! I consider [Flyte](https://docs.flyte.org/en/latest/index.html) to have a pretty well thought-out API and this is similar in style to [`FlyteFile`](https://docs.flyte.org/projects/cookbook/en/latest/auto/core/flyte_basics/files.html), so 👍

Great, thanks!

> An aside to the client-side implementation, but related to the feature itself: if we're going to support uploading potentially large files now might be the right time to implement some form of hash-based, user-scoped de-duplication to avoid scripts uploading the same file contents again and again.

@isNeil and I had a chat about this yesterday and you're dead right, we also had this in APIv1. Will write some notes and pass on.

rossgray (changes requested):
Need to remove the `.DS_Store` file

paulcjh:
> Need to remove the `.DS_Store` files
Done!

pipeline/api/cloud.py
```diff
@@ -287,17 +307,35 @@ def upload_pipeline(
...
                 io.BytesIO(python_object_to_hex(_var.type_class).encode()), "/"
             )
+
+            pipeline_file_schema: PipelineFileVariableGet = None
+
+            if isinstance(_var, PipelineFile):
+
+                _var_file_hash = self._hash_file(_var.path)
+
+                _var_file = self.upload_file(_var.path, "/")
```
rossgray, line 324: not sure how large these files are but if they could be very large then perhaps this could be made more efficient since it seems like we're reading the contents of the file twice (once to compute the hash and once in `upload_file`)?

pipeline/objects/decorators.py
```diff
@@ -54,7 +51,13 @@ def execute_func(*args, **kwargs):
             return node_output
 
     execute_func.__function__ = function
+
+    function.__run_once__ = run_once
```
rossgray, line 55: I think double underscores before and after a variable name are typically used for Python built-in magic methods (https://towardsdatascience.com/whats-the-meaning-of-single-and-double-underscores-in-python-3d27d57d6bd1).

Not sure if there's a reason we're using them specifically here rather than just something like `function._run_once`, but I see we're already doing it in a few other places, so I guess it's consistent with the current conventions we're using?

pipeline/objects/graph.py
```diff
@@ -49,39 +50,91 @@ def __init__(
...
+            node_inputs: List[Variable] = []
+            node_function: Function = None
+
+            for function in self.functions:
+                if function.local_id == node.function.local_id:
+                    node_function = function
+                    break
+            if (
+                hasattr(node_function.function, "__run_once__")
+                and node_function.function.__run_once__
```
rossgray, line 81: Think you can simplify this and the others, e.g.
```suggestion
                getattr(node_function.function, "__run_once__", False)
```

pipeline/objects/graph.py
```diff
@@ -49,39 +50,91 @@ def __init__(
...
+                        node_inputs.append(variable)
+                        break
+
+            function_inputs = []
+            for _input in node_inputs:
+                function_inputs.append(startup_variables[_input.local_id])
+
+            if node_function.function is None:
+                raise Exception(
+                    "Node function is none (id:%s)" % node.function.local_id
```
rossgray, line 102:
```suggestion
                    "Node function is None (id:%s)" % node.function.local_id
```

pipeline/objects/graph.py
```diff
@@ -49,39 +50,91 @@ def __init__(
...
+
+            if getattr(node_function, "class_instance", None) is not None:
+                node_function.function(node_function.class_instance, *function_inputs)
             else:
-                raise Exception("Model load not found")
-        self._loaded = True
+                node_function.function(*function_inputs)
+
+            if (
+                hasattr(node_function.function, "__has_run__")
```
rossgray, line 111: could be simplified too

pipeline/objects/graph.py
```diff
@@ -99,6 +152,19 @@ def run(self, *inputs):
             node_outputs: List[Variable] = []
             node_function: Function = None
 
+            for function in self.functions:
+                if function.local_id == node.function.local_id:
+                    node_function = function
+                    break
+
+            if (
+                hasattr(node_function.function, "__run_once__")
```
rossgray, line 161: these ones could also be simplified

pipeline/objects/decorators.py
```diff
@@ -1,20 +1,17 @@
...
-    """
-    Annotate a function as accesible by the Pipeline Object.
-
-        Parameters:
-                function (Function): user defined function.
-        Returns:
-                None.
-    """
+def pipeline_function(function=None, *, run_once=False, on_startup=False):
+    if function is None:
```
rossgray, line 11: maybe worth adding a comment explaining the reasoning behind this part (i.e. this is called when called with args, eg `@pipeline_function(run_once=True0`)

rossgray (commented):
Looks good! Just some minor comments.

I do wonder whether it's worth adding some tests for this though? Eg to ensure a function is really only called once. Some of these methods (eg `_startup()`) are quite complex so without some tests in place it could be easy to introduce bugs in the future

isNeil (changes requested):
some tidying up of comments while I continue read through

pipeline/objects/graph.py
paulcjh: Good suggestion! Noob mistake by me

paulcjh:
> Looks good! Just some minor comments.
> 
> I do wonder whether it's worth adding some tests for this though? Eg to ensure a function is really only called once. Some of these methods (eg `_startup()`) are quite complex so without some tests in place it could be easy to introduce bugs in the future


@rossgray I've added in two tests, one for the startup param and one for run once. I should've done this previously, thanks!

pipeline/api/cloud.py
paulcjh: Hashing a file was fast on my computer (hundreds of ms for 500MB) and the idea is to check the hash with the API in the future to check if it has been uploaded previously. Unfortunately it will mean having to open it twice.

pipeline/objects/graph.py
```diff
@@ -133,6 +182,12 @@ def run(self, *inputs):
 
             running_variables[node_outputs[0].local_id] = output
 
+            if (
+                hasattr(node_function.function, "__has_run__")
+                and not node_function.function.__has_run__
```
rossgray, line 187: this one can be simplified too

pipeline/api/cloud.py
```diff
@@ -287,17 +307,35 @@ def upload_pipeline(
 
         new_variables: List[PipelineVariableGet] = []
         print("Uploading variables")
+
+        from pipeline.objects import PipelineFile
```
rossgray, line 311: is there a reason we need to have these imports here?

pipeline/api/cloud.py
rossgray: Maybe I didn't explain well but I just meant that the `upload_file()` and `_hash_file()` are both reading in the contents of the file. So one optimisation could be to do something like (in pseudo-code):
```python
file_contents = read_file(file_path)
hash_file(file_contents)
upload_file(file_contents)
```

It might be that reading in the contents of the file is pretty fast though so maybe not worth worrying about

paulcjh: Okay gotcha - yeah I think it's best to leave this for the moment actually. The issue is the upload method for large files, imagine the file is 10GB like GPT J we don't want to load the whole thing into memory and then send. We want to send chunks of it, and the hash function doesn't load the whole thing thankfully.

pipeline/objects/decorators.py
paulcjh: Yeah you're right, I'm not sure what the original motivation was either. I did this to keep the consistency.

pipeline/objects/decorators.py
paulcjh: Great idea, done

pipeline/api/cloud.py
```diff
@@ -232,6 +245,13 @@ def upload_function(self, function: Function) -> FunctionGet:
             )
         except AttributeError as e:
             raise InvalidSchema(schema="Function", message=str(e))
+        except ValidationError as e:
+            print(
+                "Validation error when creating FunctionCreate schema for function %s."
+                % function.name
+            )
+            raise e
```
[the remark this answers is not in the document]
paulcjh: This was a thing I was using to develop something, I've removed it

pipeline/objects/graph.py
paulcjh: done

pipeline/api/cloud.py
paulcjh: Unfortunately yes, cyclical dependencies.

pipeline/api/cloud.py
rossgray: Cool, yeah we might want to update the `upload_file` method to support uploading in chunks rather than the whole file at some point too

rossgray (approved):
Thanks for the changes. Everything looks good to me

isNeil (approved):

PR #120, opened 2022-08-10
Add compute requirements to `RunGet` schema

rossgray:


pipeline/schemas/run.py
```diff
@@ -87,10 +87,20 @@ class RunGet(BaseModel):
...
         allow_population_by_field_name = True
 
+    @validator("compute_requirements", pre=True)
+    def compute_requirements_default_if_empty(cls, v):
+        """Return None if no compute_requirements rather than an empty
+        ComputeRequirements object
+        """
+        if not v:
+            return None
+        return v
```
rossgray, line 102: Not sure if this is the most Pythonic way to do this or not but seemed a bit neater to me. Basically if there are no compute requirements then we will have `schema.compute_requirements = None` rather than `schema.compute_requirements = ComputeRequirements(min_gpu_vram_mb=None)`

alexpearce: So that makes it harder (impossible?) for the Run to override the Pipeline's requirements with `None` attributes? e.g. if the Pipeline specifies 8 GB but the Run wants to override that with no requirement.

Probably not a use-case worth spending much time on for now anyway, just wanted to see if I understand the logic.

alexpearce (approved):

pipeline/schemas/run.py
rossgray: I don't believe it should affect that logic since the overriding logic is handled [here](https://github.com/mystic-ai/pipeline-top/blob/main/app/services/run.py#L90-L92) and just takes into account the `RunCreate` data and the Pipeline's requirements from the DB.

At a high level the way I understand it is that the `RunCreate` specifies requirements for creating the run. These are then combined with the requirements on the Pipeline and then a `Run` is created in the DB (a `PipelineRun` or `FunctionRun`). The `RunGet` schema then represents the actual `Run` we have created.

The intention behind this change is to allow us to just to pass `None` to the cluster controller if there are no overall compute requirements rather than an empty `ComputeRequirements` object (my thinking is that it is neater this way since the controller immediately knows there are no compute requirements rather than checking each field in the `ComputeRequirements` to see if they're `None`). You can see this in this test [here](https://github.com/mystic-ai/pipeline-top/pull/242/files#diff-8d185ecb4dd3eca5458bd56b7f1eff971234e31739ac8cf6f6218ba60d7eb129R472).

(Another way to do the above would be to add a method to the `ComputeRequirements` to say whether it is effectively empty or not, or add a `__bool__` method (but not sure if that would mess with Pydantic's internals)).

Let me know if any of this doesn't make sense or if some of my assumptions are wrong.

alexpearce: Makes total sense, yes! I missed that this is on the `RunGet` schema rather than `RunCreate`.

PR #121, opened 2022-08-15
Plutopulp/username validator update

plutopulp:
This PR fixes a bug in the `UserCreate` schemas' `username_validator`, which previously raised an error for usernames with the value of `None`. 
I had a scan through other validators to see if a similar issue was present, but didn't see any.
Note, the test seems a bit strange, there may be a better way of checking that optional usernames are in fact valid. If anyone has a better idea for a test, I'd appreciate your input.

### DIFF
--- a/tests/schemas/test_user.py
+++ b/tests/schemas/test_user.py
@@ -1,0 +1,9 @@
+from pipeline.schemas.user import UserCreate
+
+
+def test_user_create_optional_username():
+    none_username_payload = dict(
+        email="email@company.com", password="ExamplePass123", username=None
+    )
+    UserCreate(**none_username_payload)
+    assert True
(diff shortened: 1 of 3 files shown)


tests/schemas/test_user.py
```diff
@@ -0,0 +1,9 @@
+from pipeline.schemas.user import UserCreate
+
+
+def test_user_create_optional_username():
+    none_username_payload = dict(
+        email="email@company.com", password="ExamplePass123", username=None
+    )
+    UserCreate(**none_username_payload)
+    assert True
```
rossgray, line 9: Don't think you really need this line since it's not really doing anything.

Not sure if we already have some other tests for the validator but might be worth adding one test for a valid username and one test for an invalid one if it's quick to do?

plutopulp: Thanks man! Removed the `assert True`and added some tests for valid + invalid usernames.

rossgray (approved):
LGTM 🚢

PR #124, opened 2022-08-22
Fix matrix dimensions in eg

brendanbennett:
Fixed an example script failing to run because of incorrect matrix dimensions.


brendanbennett:
This doesn't really need a version bump. Can be put in next version.

rossgray (approved):
LGTM

PR #130, opened 2022-08-26
Oscar/onnx add name

hirovi:


brendanbennett (approved):
Approved courtesy of @isNeil

PR #131, opened 2022-08-31
Direct PipelineFile uploads

rossgray:
Allow client to upload files directly. This is needed for `PipelineFile`s which can be very large.

TODO:
- [x] Add tests

### DIFF
--- a/pipeline/api/cloud.py
+++ b/pipeline/api/cloud.py
@@ -28,13 +28,13 @@ from pipeline.schemas.pipeline import (
     PipelineVariableGet,
 )
 from pipeline.schemas.pipeline_file import (
+    MultipartUploadMetadata,
     PipelineFileDirectUploadFinaliseCreate,
     PipelineFileDirectUploadInitCreate,
     PipelineFileDirectUploadInitGet,
     PipelineFileDirectUploadPartCreate,
     PipelineFileDirectUploadPartGet,
     PipelineFileGet,
-    MultipartUploadMetadata,
 )
 from pipeline.schemas.run import RunCreate
 from pipeline.util import (
--- a/pipeline/api/cloud.py
+++ b/pipeline/api/cloud.py
@@ -18,7 +18,12 @@ from pipeline.exceptions.MissingActiveToken import MissingActiveToken
 from pipeline.schemas.base import BaseModel
 from pipeline.schemas.compute_requirements import ComputeRequirements
 from pipeline.schemas.data import DataGet
-from pipeline.schemas.file import FileCreate, FileGet
+from pipeline.schemas.file import (
+    FileCreate,
+    FileGet,
+    FileDirectUploadCreate,
+    FileDirectUploadGet,
+)
 from pipeline.schemas.function import FunctionCreate, FunctionGet
 from pipeline.schemas.model import ModelCreate, ModelGet
 from pipeline.schemas.pipeline import (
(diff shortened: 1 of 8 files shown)


pipeline/schemas/file.py
```diff
@@ -12,7 +12,7 @@ class FileGet(FileBase):
     path: str
     #: The data as hex-encoded bytes, if the data size is less than 20 kB
     data: Optional[str]
-    #: The data size in kilobytes (kB)
+    #: The data size in bytes
```
alexpearce, line 15: Thank you! 🙏

pipeline/api/cloud.py
```diff
@@ -145,6 +155,103 @@ def upload_python_object_to_file(self, obj, remote_path) -> FileGet:
...
+            multipart_metadata=multipart_metadata,
+        )
+        response = self._post(
+            "/v2/pipeline-files/finalise-multipart-upload",
+            finalise_upload_schema.dict(),
+        )
+        return PipelineFileGet.parse_obj(response)
+
+    def upload_pipeline_file(self, pipeline_file) -> PipelineFileVariableGet:
+        """Upload PipelineFile given by pipeline_file.
```
alexpearce, line 205: Can we type `pipeline_file`? Then the docstring can be a bit more to the point.

```suggestion
    def upload_pipeline_file(self, pipeline_file: PipelineFile) -> PipelineFileVariableGet:
        """Upload the pipeline input file.
```

pipeline/api/cloud.py
```diff
@@ -145,6 +155,103 @@ def upload_python_object_to_file(self, obj, remote_path) -> FileGet:
...
+
+        Since PipelineFiles can be very large, we implement this slightly
+        differently to regular file uploads:
+        - We need to split the file into chunks based on FILE_CHUNK_SIZE
+        - We first initialise the multi-part upload with the server
+        - We then upload the file in chunks (requesting a presigned upload URL for each
+            chunk beforehand)
+        - Lastly, we finalise the multi-part upload with the server
```
alexpearce, line 213: Great docs!

pipeline/api/cloud.py
```diff
@@ -145,6 +155,103 @@ def upload_python_object_to_file(self, obj, remote_path) -> FileGet:
...
+        response = self._post(
+            "/v2/pipeline-files/presigned-url", part_upload_schema.dict()
+        )
+        part_upload_get = PipelineFileDirectUploadPartGet.parse_obj(response)
+        # upload file chunk
+        # convert data to hex
+        data = data.hex().encode()
+        response = requests.put(part_upload_get.upload_url, data=data)
+        etag = response.headers["ETag"]
+        return {"ETag": etag, "PartNumber": part_num}
```
alexpearce, line 188: Reckon it's worth making a little schema for this? Then `PipelineFileDirectUploadFinaliseCreate.multipart_metadata` can be typed using that rather than `List[dict]`.

alexpearce (changes requested):
Just a few nits from me. Clean and compact diff 💯

pipeline/api/cloud.py
rossgray: I wanted to but I don't think we can due to circular dependencies (i.e. if I use a type annotation I have to import it at the module level and this breaks things). Would be really good to fix these circular imports at some point

pipeline/api/cloud.py
rossgray: good idea!

pipeline/api/cloud.py
rossgray: Yeah but there's an `import` just above it in the same method

pipeline/api/cloud.py
rossgray: done now

pipeline/api/cloud.py
alexpearce: D'oh, I'm blind. Ignore me!

alexpearce (approved):
:shipit:

PR #136, opened 2022-09-29
add optional start+end time to compute schema, and add total compute …

plutopulp:
This PR adds start and end times in the response schema of pipeline compute. 
Also introduces a new response schema for total compute.


isNeil (approved):
LGTM

PR #137, opened 2022-10-03
Fix to getting source code from inspect in notebooks

paulcjh:
Notebook environments aren't always compatible with the `inspect.getsource(...)` function. When this is the case we now record a random UUID so that the hash generated for the model is random.

### DIFF
--- a/pipeline/objects/function.py
+++ b/pipeline/objects/function.py
@@ -1,3 +1,5 @@
+import uuid
+
 import inspect
 from hashlib import sha256
 from typing import Any, Callable, Dict, Optional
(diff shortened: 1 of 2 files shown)


rossgray:
I'm not too familiar with this area of the codebase - What was the original purpose of using `inspect.getsource()`? Is it purely to generate a hash for the function/model? Just trying to understand why we can't always use a random UUID for example. Do we use the hash of the actual source code for caching or something?

paulcjh:
> I'm not too familiar with this area of the codebase - What was the original purpose of using `inspect.getsource()`? Is it purely to generate a hash for the function/model? Just trying to understand why we can't always use a random UUID for example. Do we use the hash of the actual source code for caching or something?

It was for display on the dashboard. There is scope to use the hash as to not re-upload, but not at the moment.

rossgray (approved):
Cool. Changes look good to me then.

As an aside, I think it would be good to document the various attributes of these objects so it's clear what they're being used for

PR #143, opened 2022-10-06
Oscar/add_io_file_descriptor

hirovi:
This PR adds a file handler designed for runtime. Used in pipelines where the input requires a local file not found on the server. Example: Whisper model by OpenAI expects the audio file that needs to be processed, this PR adds helper functions to serialise the audio file.

### DIFF
--- a/pipeline/io.py
+++ b/pipeline/io.py
@@ -33,8 +33,8 @@ class File():
         file_ext = file_as_str_with_extension[file_as_str_with_extension.find(self.key)+len(self.key):file_as_str_with_extension.rfind(self.key)]
         return cleaned_string, file_ext
 
-    def file_str_to_b64(self) -> Union[bytes, list]:
-        cleaned_str, file_ext = self._remove_ext_from_string()
+    def file_str_to_b64(self, file_as_str_with_extension: str) -> Union[bytes, list]:
+        cleaned_str, file_ext = self._remove_ext_from_string(file_as_str_with_extension)
         file_as_b64 = base64.decodebytes(cleaned_str.encode())
         return file_as_b64, file_ext
     
--- a/pipeline/io.py
+++ b/pipeline/io.py
@@ -1,0 +1,45 @@
+import base64
+from typing import Union
+
+class File():
+    def __init__(self, path: str) -> None:
+        self.path = path
+        self._key = "ext_file" # Used to append the extension to the file string and extract it during pipeline execution
+        pass
+
+    def _file_to_str_with_extension(self, file_path: str) -> str:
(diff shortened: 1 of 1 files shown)


pipeline/io.py
```diff
@@ -0,0 +1,99 @@
...
+
+        file = File()
+        file_as_bytes = file.file_get_bytes(input_data)
+        temp_file = tempfile.NamedTemporaryFile(delete=True, mode="wb", suffix=file.extension, dir="")
+        temp_file.write(file_as_bytes)
+
+        # Do whatever to the file
+        # ...
+        # Close and delete file
+        temp_file.close()
```
alexpearce, line 37: Am I right in thinking that the user should use code like this inside their Pipeline functions?

alexpearce (commented):
Thank you for this!

The general concept seems to be to serialise an input file on the user's machine as a string, and pass that string as a 'regular' input to their pipeline. The string should then be written back to a file when the pipeline is run. Is that right?

pipeline/io.py
hirovi: That's right! In the case of Whisper's model, the input expected is an audio file (hence the example with a temp file). The user is expected to re-use this File class inside pipeline function to manipulate the file in any desired format.

pipeline/io.py
```diff
@@ -0,0 +1,99 @@
...
+
+    Example: (Transfer an audio file)
+
+        '''
+        On client
+        '''
+
+        from pipeline.io import File
+
+        FILE_PATH = "files/audio.flac"
```
alexpearce, line 16: Having to encode the extension leads to a bit of a dance. Could we skip trying to ship the extension entirely? The user's function anyhow has to do the `File.file_get_bytes` call, so if they really need the extension they can include it again there.

paulcjh:
A few thoughts:
- It would be nice if there was a more organised placement of the`io.py` file or its objects, perhaps in the variables directory?
- The naming of some of the functions I found quite confusing, for example the `to_api()`. At first, I thought it was actually uploading to the API but when I looked through the source I saw it's just invoking the inner function `_file_to_str_with_extension()`.
- We have an existing system to turn an arbitrary object into a string, https://github.com/mystic-ai/pipeline/blob/main/pipeline/util/__init__.py, I would suggest using that over the base64 encode -> decode system used
- I think, as a user, this will be quite confusing as the "pipeline" that this would be used with in the example would actually expect a string as the input, but here we're telling users to use a `File` object not a string.
- A full pipeline example would be useful for this

IMO a util function would be the optimal solution here instead of a full new `io` object.

PR #144, opened 2022-10-10
Exception on upload run

paulcjh:
Pipelines that have been run locally cannot be correctly uploaded due to serialisation issues. This PR raises an exception when a user tries to upload a pipeline that has already been run locally.


isNeil (approved):
lgtm

PR #145, opened 2022-10-11
Asyncio API runs 

paulcjh:
The `requests` library is used for the `run_pipeline(...)` function which makes it unusable for applications that utilise asyncio. 
This PR adds asyncio compatibility via `pipeline.api.asyncio.PipelineCloud` which is interchangeable with `pipeline.PipelineCloud`. Also, as this is targeted towards production API usage, all verbose information reporting has been removed including:

- Initial authentication, it's now per request
- Print statements
- Progress bars

Addition changes:
- An optional `verbose` setting for `PipelineCloud` which removes print outs + progress bars
- The `RunGet` schema is also return from `PipelineCloud.run_pipeline(...)`
- The `token` and `url` fields must be used explicitly as named kwargs now, this resolves some confusion expressed by users


examples/asyncio/stable_diffusion_run.py
```diff
@@ -0,0 +1,16 @@
+import asyncio
+
+from pipeline.api.asyncio import PipelineCloud
+
+api = PipelineCloud()
+
+
+async def run():
+    res = await api.run_pipeline(
+        "pipeline_67d9d8ec36d54c148c70df1f404b0369", [["a dog"], {}]
```
rossgray, line 10: might be better to use a pipeline alias here?

paulcjh: How do I do that?

pipeline/api/cloud.py
```diff
@@ -531,7 +531,8 @@ def run_pipeline(
             compute_type=compute_type,
             compute_requirements=compute_requirements,
         )
-        return self._post("/v2/runs", json.loads(run_create_schema.json()))
+        run_json: dict = self._post("/v2/runs", json.loads(run_create_schema.json()))
+        return RunGet.parse_obj(run_json)
```
rossgray, line 535: This would be a breaking change in the API (i.e. clients would need to update their code when upgrading to this version). Therefore, if we want to go ahead with this we would need to ensure we make this a major update (e.g bump version to 1.0.0)

rossgray:
Just some general comments from me:

There is quite a bit of code duplication here (between `pipeline/api/asyncio/cloud.py` and `pipeline/api/cloud.py`). Some of this may be unavoidable due to async vs non-async but wonder if some could be shared? Otherwise, we probably want to make sure we keep both versions of the client consistent with one another (for example to make it easy to switch between them and to make maintenance easier. A good example here is `httpx` and `requests` which have almost identical interfaces). Therefore, it might be worth adding some tests for this since it's probably going to be easy in the future to update the regular client and forget to update the async one, or vice-versa.

> Also, as this is targeted towards production API usage, all verbose information reporting has been removed including
> - Initial authentication, it's now per request
> - Print statements
> - Progress bars

Again, while these are good ideas, I think it would be good to keep both the async and non-async client in sync here. I think either could genuinely be used in a production environment, so maybe the same changes could be made to the regular client? (I would argue some things like the progress bars are useful for local development or people using the library for the first time, so maybe we could implement those using logging instead of print statements? That way they can be ignored in production code but kept in other cases. See https://docs.python-guide.org/writing/logging/#logging-in-a-library)

examples/asyncio/stable_diffusion_run.py
rossgray: These need to be setup on the backend but you should be able to pass in an alias instead of the pipeline id. I'm not sure if any have been set up yet so might be best to check with @alexpearce

rossgray (changes requested):
Just some general comments from me:

There is quite a bit of code duplication here (between pipeline/api/asyncio/cloud.py and pipeline/api/cloud.py). Some of this may be unavoidable due to async vs non-async but wonder if some could be shared? Otherwise, we probably want to make sure we keep both versions of the client consistent with one another (for example to make it easy to switch between them and to make maintenance easier. A good example here is httpx and requests which have almost identical interfaces). Therefore, it might be worth adding some tests for this since it's probably going to be easy in the future to update the regular client and forget to update the async one, or vice-versa.

> Also, as this is targeted towards production API usage, all verbose information reporting has been removed including
> 
> Initial authentication, it's now per request
> Print statements
> Progress bars

Again, while these are good ideas, I think it would be good to keep both the async and non-async client in sync here. I think either could genuinely be used in a production environment, so maybe the same changes could be made to the regular client? (I would argue some things like the progress bars are useful for local development or people using the library for the first time, so maybe we could implement those using logging instead of print statements? That way they can be ignored in production code but kept in other cases. See https://docs.python-guide.org/writing/logging/#logging-in-a-library)

examples/asyncio/stable_diffusion_run.py
paulcjh: Okay, I think resolving this in a separate PR where we bring all of the examples forward to the new alias system might be best?

pipeline/api/cloud.py
paulcjh: Just did a VB to 0.1.0

paulcjh:
> Just some general comments from me:
> 
> There is quite a bit of code duplication here (between pipeline/api/asyncio/cloud.py and pipeline/api/cloud.py). Some of this may be unavoidable due to async vs non-async but wonder if some could be shared? Otherwise, we probably want to make sure we keep both versions of the client consistent with one another (for example to make it easy to switch between them and to make maintenance easier. A good example here is httpx and requests which have almost identical interfaces). Therefore, it might be worth adding some tests for this since it's probably going to be easy in the future to update the regular client and forget to update the async one, or vice-versa.

For sure, there is a lot of duplication. There was no clear path that I could see that allowed the two PipelineCloud versions to be interchangeable and not replicate code. I based the interchangeable nature on a few other libs I saw that do the same. I agree on the tests, I'll take a look at that now.

> 
> > Also, as this is targeted towards production API usage, all verbose information reporting has been removed including
> > Initial authentication, it's now per request
> > Print statements
> > Progress bars
> 
> Again, while these are good ideas, I think it would be good to keep both the async and non-async client in sync here. I think either could genuinely be used in a production environment, so maybe the same changes could be made to the regular client? (I would argue some things like the progress bars are useful for local development or people using the library for the first time, so maybe we could implement those using logging instead of print statements? That way they can be ignored in production code but kept in other cases. See https://docs.python-guide.org/writing/logging/#logging-in-a-library)

We've had the removal of logs requested on the regular client, so we can add an option for that. The main distinction is that the asyncio package will **only** be used in a production env, but as you rightfully point out the logs are useful for development in the regular client.

rossgray:
I wonder if a better way to reduce duplication would be for the async client to inherit from the regular client. Then they can share the same helper methods, etc and help reduce duplication. For example:

```python
import asyncio


class Client:
    def helper(self):
        print("helping")

    def post(self):
        self.helper()
        print("Sync post")


class AsyncClient(Client):
    async def post(self):
        self.helper()
        print("Async post")


async def main():
    client = Client()
    client.post()
    client = AsyncClient()
    await client.post()


asyncio.run(main())
```

pipeline/api/cloud.py
rossgray: I think _strictly_ this should be a bump to 1.0.0 (see https://semver.org/)

examples/asyncio/stable_diffusion_run.py
rossgray: yeah can do

pipeline/api/cloud.py
paulcjh: Okay!

paulcjh:
> I wonder if a better way to reduce duplication would be for the async client to inherit from the regular client. Then they can share the same helper methods, etc and help reduce duplication. For example:
> 
> ```python
> import asyncio
> 
> 
> class Client:
>     def helper(self):
>         print("helping")
> 
>     def post(self):
>         self.helper()
>         print("Sync post")
> 
> 
> class AsyncClient(Client):
>     async def post(self):
>         self.helper()
>         print("Async post")
> 
> 
> async def main():
>     client = Client()
>     client.post()
>     client = AsyncClient()
>     await client.post()
> 
> 
> asyncio.run(main())
> ```

Great idea, on it

pipeline/api/asyncio/cloud.py
```diff
@@ -0,0 +1,242 @@
...
+        if min_gpu_vram_mb:
+            compute_requirements = ComputeRequirements(min_gpu_vram_mb=min_gpu_vram_mb)
+
+        run_create_schema = RunCreate(
+            pipeline_id=pipeline_id,
+            data_id=_data_id,
+            compute_type=compute_type,
+            compute_requirements=compute_requirements,
+        )
+        run_json: dict = self._post("/v2/runs", json.loads(run_create_schema.json()))
```
rossgray, line 241:
```suggestion
        run_json: dict = await self._post("/v2/runs", json.loads(run_create_schema.json()))
```

pyproject.toml
```diff
@@ -1,6 +1,7 @@
 [tool.poetry]
 name = "pipeline-ai"
-version = "0.0.49"
+version = "1.0.0"
```
rossgray, line 3: sorry, my bad, @alexpearce pointed out that if we're not yet on version 1 we can still introduce breaking changes

pipeline/api/cloud.py
```diff
@@ -55,9 +55,19 @@ class PipelineCloud:
...
 
-    def __init__(self, url: str = None, token: str = None) -> None:
+    def __init__(
+        self,
+        *,
+        url: str = None,
+        token: str = None,
+        timeout=60.0,
+        verbose=True,
+        auth_check=True,
```
rossgray, line 65: not sure what the intention of this is - don't think it's currently used anywhere?

pipeline/api/cloud.py
```diff
@@ -71,7 +81,9 @@ def authenticate(self, token: str = None):
             Returns:
                 None
         """
-        print("Authenticating")
+        if self.verbose:
```
rossgray, line 84: nice

pipeline/api/asyncio/cloud.py
```diff
@@ -0,0 +1,242 @@
...
+    def download_result(self, result_id_or_schema: Union[str, RunGet]) -> Any:
+        self._raise_not_implemeneted()
+
+    def download_pipeline(self, id: str) -> Graph:
+        self._raise_not_implemeneted()
+
+    @staticmethod
+    def _get_raise_for_status(response: httpx.Response) -> None:
+        # A handler for errors that might be sent with messages from Top.
+        if not response.status_code == httpx.codes.OK:
```
rossgray, line 95: there's still quite a bit of code duplication across the async and sync client, for example this method is identical to the sync version apart from this one line. Maybe it's worth making this one line work with both `httpx` and `requests`? Or we could just switch to `httpx` entirely since it supports both sync and async?

@alexpearce - any opinions on this?

pipeline/api/cloud.py
```diff
@@ -55,9 +55,19 @@ class PipelineCloud:
     token: Optional[str]
     url: Optional[str]
 
-    def __init__(self, url: str = None, token: str = None) -> None:
+    def __init__(
+        self,
+        *,
+        url: str = None,
+        token: str = None,
+        timeout=60.0,
```
rossgray, line 63: it's probably worth adding timeout to the sync client methods too to make things consistent, otherwise this is probably going to confuse the end user if they're specifying the timeout and it does nothing

pipeline/api/asyncio/cloud.py
```diff
@@ -0,0 +1,242 @@
...
+    def _get(self, endpoint: str, params: Dict[str, Any] = None):
+        self._raise_not_implemeneted()
+
+    def upload_function(self, function: Function) -> FunctionGet:
+        self._raise_not_implemeneted()
+
+    def upload_model(self, model: Model) -> ModelGet:
+        self._raise_not_implemeneted()
+
+    def upload_pipeline(
```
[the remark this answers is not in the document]
paulcjh: We can add these in with follow up PRs

pyproject.toml
paulcjh: Okay

pipeline/api/cloud.py
paulcjh: Correct, removed.

pipeline/api/cloud.py
paulcjh: I made a note of this before, done now

pipeline/api/asyncio/cloud.py
paulcjh: I've added in something to fix this using a union typing input and checking if the response is httpx or requests. In a follow up PR we can consolidate the library used, no need at the moment I don't think.

pipeline/api/asyncio/cloud.py
```diff
@@ -0,0 +1,220 @@
...
+        compute_requirements = None
+        if min_gpu_vram_mb:
+            compute_requirements = ComputeRequirements(min_gpu_vram_mb=min_gpu_vram_mb)
+
+        run_create_schema = RunCreate(
+            pipeline_id=pipeline_id,
+            data_id=_data_id,
+            compute_type=compute_type,
+            compute_requirements=compute_requirements,
+        )
```
rossgray, line 216: I think it would still be good to reduce some of this code duplication but we can do that in a follow up PR

rossgray (commented):
I think this is fine for an initial version. I think it would be worth cleaning some of the code up, implementing the other methods and adding more tests, but this can be done in a follow up PR.

I would just change the version to a minor release so people know the interface has changed slightly.

rossgray (approved):
LGTM

PR #146, opened 2022-10-11
Spacy helper

isNeil:


examples/spacy/spacy-tokenizer.py
```diff
@@ -0,0 +1,66 @@
...
+
+    @pipeline_function
+    def predict(self, input: str) -> list:
+        doc = self.nlp(input)
+        # (optional) your spacy code here or you can return entire spacy object
+        # to manipulate on a client that has spacy installed.
+        res = []
+        for token in doc:
+            res.append([token.text, token.pos_, token.dep_])
+        return res
```
rossgray, line 23: just a style thing so not important, but I quite like using list comprehensions for this kind of thing, e.g.
```python
return [[token.text, token.pos_, token.dep_] for token in doc]

Comment (alexpearce) [inline examples/spacy/spacy-tokenizer.py:23]:
```suggestion
        return [[token.text, token.pos_, token.dep_] for token in doc]
```

alexpearce (commented):
Just one suggestion for your consideration, otherwise LGTM!

I guess we don't need to bump the version when adding examples but no harm either.

rossgray (approved):
Looks good to me

alexpearce (approved):
🚢

PR #147, opened 2022-10-11
Add more detailed error info in RunGet

rossgray:
Extra error info added so that exceptions and tracebacks can be reported to the client. Changes to the schemas have been made in a backwards-compatible way.


alexpearce (approved):
LGTM!

> Changes to the schemas have been made in a backwards-compatible way.

Thank you!

PR #148, opened 2022-10-12
Remove harmful replace call

alexpearce:
This blindly modifies the pickle payload, potentially invalidating it.


paulcjh (commented):
Do we know why this was originally added in? Seems really peculiar.

alexpearce:
> Do we know why this was originally added in?

It was added in https://github.com/mystic-ai/pipeline/pull/107 as part of a beta release to test a failure seen in an internal service. The beta release did not help that failure but the change was not reverted, so it stayed for subsequent releases.

isNeil (approved):
lgtm

PR #152, opened 2022-10-19
Improve progress feedback for pipeline file uploads

rossgray:
These changes wrap the data passed to `requests.post` in a `BytesIO` object, which is in turn wrapped in a `tqdm`'s `CallbackIOWrapper`. This allows `requests` to read in the data incrementally, which in turn calls the `tqdm` callback to update the progress bar.

Inspiration for this was taken from this [gist](https://gist.github.com/tyhoff/b757e6af83c1fd2b7b83057adf02c139)

### DIFF
--- a/pipeline/api/cloud.py
+++ b/pipeline/api/cloud.py
@@ -13,6 +13,7 @@ import requests
 from pydantic import ValidationError
 from requests_toolbelt.multipart import encoder
 from tqdm import tqdm
+from tqdm.utils import CallbackIOWrapper
 
 from pipeline.exceptions.InvalidSchema import InvalidSchema
 from pipeline.exceptions.MissingActiveToken import MissingActiveToken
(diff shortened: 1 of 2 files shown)


alexpearce (dismissed):
Very cool!

Is there already test coverage for this code?

Is there a need to add similar functionality to the asyncio implementation? Or was it agreed not to support logging there?

rossgray:
> Is there already test coverage for this code?

There is the following [test](https://github.com/mystic-ai/pipeline/blob/main/tests/test_pipeline_cloud.py#L73-L83) for the pipeline file upload

> Is there a need to add similar functionality to the asyncio implementation? Or was it agreed not to support logging there?

The async client doesn't implement most of the methods on the regular client currently, including pipeline file uploads. I think it currently only handles running of pipelines and not uploading.

rossgray:
sorry @alexpearce I needed to bump the version

alexpearce (approved):

PR #156, opened 2022-10-25
add warning on pipeline upload from non-3.9 env

65536william:
Uploading pipelines in 3.10 and occasionally in 3.8 is buggy and often mysteriously so. This change at least advises users to use Python 3.9, until upload support for 3.10 and onwards is in place.

### DIFF
--- a/pipeline/api/cloud.py
+++ b/pipeline/api/cloud.py
@@ -442,8 +442,7 @@ class PipelineCloud:
                 " upload_pipeline function has only been tested in Python 3.9. "
                 "We strongly recommend you use Python 3.9 as pipelines uploaded"
                 " in other Python versions are known to be broken. We are working"
-                "on adding support for 3.10 and 3.8; join us on Discord"
-                " https://discord.com/invite/7REbAX5v3N if you'd like to help us!"
+                "on adding support for 3.10 and 3.8!"
             )
 
         new_name = new_pipeline_graph.name
--- a/pipeline/api/cloud.py
+++ b/pipeline/api/cloud.py
@@ -435,9 +435,8 @@ class PipelineCloud:
         """
         if new_pipeline_graph._has_run_startup:
             raise Exception("Cannot upload a pipeline that has already been run.")
-        if sys.version_info.major != 3 or (
-            sys.version_info.major == 3 and sys.version_info.minor != 9
-        ):
+        # Pipeline Cloud currently supports Python 3.9.x only
+        if (sys.version_info.major, sys.version_info.minor) != (3, 9):
             print(
                 "WARNING: pipeline-ai is still in development and the"
                 " upload_pipeline function has only been tested in Python 3.9. "
--- a/pipeline/api/cloud.py
+++ b/pipeline/api/cloud.py
@@ -7,6 +7,7 @@ import os
 import urllib.parse
 from http import HTTPStatus
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Type, Union
+import sys
 
 import httpx
 import requests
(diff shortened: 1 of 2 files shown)


pipeline/api/cloud.py
```diff
@@ -434,6 +435,17 @@ def upload_pipeline(
...
+        if sys.version_info.major != 3 or (
+            sys.version_info.major == 3 and sys.version_info.minor != 9
+        ):
+            print(
+                "WARNING: pipeline-ai is still in development and the"
+                " upload_pipeline function has only been tested in Python 3.9. "
+                "We strongly recommend you use Python 3.9 as pipelines uploaded"
+                " in other Python versions are known to be broken. We are working"
+                "on adding support for 3.10 and 3.8; join us on Discord"
+                " https://discord.com/invite/7REbAX5v3N if you'd like to help us!"
```
alexpearce, line 447: I don't think we can make this call-to-action as it likely requires work on the backend.

pipeline/api/cloud.py
```diff
@@ -434,6 +435,17 @@ def upload_pipeline(
         """
         if new_pipeline_graph._has_run_startup:
             raise Exception("Cannot upload a pipeline that has already been run.")
+        if sys.version_info.major != 3 or (
+            sys.version_info.major == 3 and sys.version_info.minor != 9
+        ):
+            print(
```
alexpearce, line 441: If we're certain that non-3.9 pipelines will never run we should stop execution here, e.g. raise an exception.

pipeline/api/cloud.py
```diff
@@ -434,6 +435,17 @@ def upload_pipeline(
         """
         if new_pipeline_graph._has_run_startup:
             raise Exception("Cannot upload a pipeline that has already been run.")
+        if sys.version_info.major != 3 or (
+            sys.version_info.major == 3 and sys.version_info.minor != 9
+        ):
```
alexpearce, line 440:
```suggestion
        # Pipeline Cloud currently supports Python 3.9.x only
        if (sys.version_info.major, sys.version_info.minor) != (3, 9):
```

alexpearce (commented):
Thanks for this, it'll be really helpful for folks to see this message upfront.

How about putting a notice in the README, maybe at the beginning of the [usage section](https://github.com/mystic-ai/pipeline#usage)?

pipeline/api/cloud.py
65536william: nice tuple moment

pipeline/api/cloud.py
65536william: In the past some 3.8 pipelines have worked for me which is why I chose `print` vs `raise`. Ideally actually we'd be using the `logger` stuff so it could be a proper debugging warning.

pipeline/api/cloud.py
65536william: The Mystic jobs page no longer exists, so Oscar and I figured send them to Discord and go from there (but with the aim that somebody seriously keen will speak to us and potentially be hired). Do you think I should clarify, i.e. "if you'd like to work with us"?

pipeline/api/cloud.py
alexpearce: I see. I'd still be tempted to raise as running this with non-3.9 is effectively undefined behaviour, but leave that up to you.

pipeline/api/cloud.py
alexpearce: I don't think this is appropriate place to try to hire people but leave that up to you and Oscar 😄

alexpearce (approved):
This'll be really helpful! And thanking for cleaning up the whitespace in the README 💯

PR #163, opened 2022-11-17
Paul/cleanup

paulcjh:
PR Checklist:
- [ ] Docs updated
- [ ] Tests written

### DIFF
--- a/examples/docker/README.md
+++ b/examples/docker/README.md
@@ -1,0 +1,103 @@
+# Auto-docker API with Pipeline
+
+Pipline has support for creating FastAPI endpoints for your Pipelines out of the box. To demonstrate this we'll create a basic GPT-Neo Pipeline with a custom environment and deploy the docker image.
+
+Required to run the code below (and all auto-docker features in Pipeline) an uptodate installation of docker, and docker compose, is required.
+
+## API architecture
+
+The auto-docker feature from the Pipeline library extends the base docker image found here: [pipeline-docker](https://github.com/mystic-ai/pipeline-docker).
+
+There are three main Docker images that will be added into the final docker compose that are required to run the API:
+1. `pipeline-docker` - The main image containing the API.
+2. `postgresql` - This is used for traffic logging, and will be extended to include auth.
+3. `redis` - For session data.
+
+## GPT Neo example
+### Pipeline creation
+
+We'll create a simple pipeline to take in an input string, and run it through GPT-Neo. A primitive caching method is implementing when on the first run the model and tokenizer are created and stored in the main model class.
+
+```python
+from pipeline import Pipeline, Variable, pipeline_model, pipeline_function
+
+@pipeline_model
+class GPTNeo:
+    def __init__(self):
+        self.model = None
+        self.tokenizer = None
+
+    @pipeline_function
+    def predict(self, input_data: str) -> str:
+        if self.model == None:
+            # Import the transformers modules here so that they'represent when
+            # executed in a seperate env post serialisation.
+            from transformers import GPTNeoForCausalLM, GPT2Tokenizer
+            self.model = GPTNeoForCausalLM.from_pretrained("EleutherAI/gpt-neo-125M")
+            self.tokenizer = GPT2Tokenizer.from_pretrained("EleutherAI/gpt-neo-125M")
+
+        input_ids = self.tokenizer(input_data, return_tensors="pt").input_ids
+
+        gen_tokens = self.model.generate(
+            input_ids,
+            do_sample=True,
+            temperature=0.9,
+            max_length=100,
+        )
+        gen_text = self.tokenizer.batch_decode(gen_tokens)[0]
+        return gen_text
+
+# Create the computation graph using the Pipeline context manager
+with Pipeline("gptneo") as builder:
+    in_1 = Variable(str, is_input=True)
+    builder.add_variables(in_1)
+
+    gpt_neo = GPTNeo()
+
+    out_str = gpt_neo.predict(in_1)
+
+    builder.output(out_str)
+
+gptneo_pl = Pipeline.get_pipeline("gptneo")
+```
+
+### Environment creation
+Now that we have created the pipeline we can and create the environment that we'll need to run it.
+
+```python
+from pipeline.objects.environment import Environment
+
+# An environment takes in a list of dependencies, in our case we only require transformers and torch.
+# Note: You can pass in any generic dependency string (including git repos).
+pl_env = Environment(
+    "gptneo-env",
+    dependencies=[
+        "transformers==4.24.0",
+        "torch==1.13.0"
+    ],
+)
+```
+
+### Auto Docker
+Finally, using the environment and pipeline graph we can create the docker files.
+
+```python
+from pipeline import docker
+
+docker.create_pipeline_api([gptneo_pl], environment=pl_env)
+```
+
+Running this final command will generate 4 files in your local directory:
+1. `docker-compose.yml` - A docker compose file containing environment variables, build info etc.
+2. `Dockerfile` - The Dockerfile that is used to create the final container
+3. `gptneo.graph` - The serialised graph createed by the Pipeline context manager
+4. `requirements.txt` - The python dependencies as defined above (`["transformers==4.24.0","torch==1.13.0"]`)
+
+### Using the auto-docker API
+
+
+## Not supported
+
+Currently the following Variables are not supported with the auto-docker feature:
+- PipelineFile
+- Local directories as dependencies for the environment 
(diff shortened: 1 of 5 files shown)


examples/docker/README.md
```diff
@@ -0,0 +1,123 @@
+# Auto-docker API with Pipeline
+
+Pipline has support for creating FastAPI endpoints for your Pipelines out of the box. To demonstrate this we'll create a basic GPT-Neo Pipeline with a custom environment and deploy the docker image.
```
rossgray, line 3:
```suggestion
Pipeline has support for creating FastAPI endpoints for your Pipelines out of the box. To demonstrate this we'll create a basic GPT-Neo Pipeline with a custom environment and deploy the docker image.
```

examples/docker/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+Required to run the code below (and all auto-docker features in Pipeline) an uptodate installation of docker, and docker compose, is required.
```
rossgray, line 5:
```suggestion
To run the code below (and all auto-docker features in Pipeline) an up-to-date installation of docker, and docker compose, is required.
```

pipeline/objects/environment/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+            self.merge_with_environment(_env)
+
+    def to_requirements(self, output_dir: str = "./") -> None:
+        requirements_path = os.path.join(os.path.join(output_dir, "requirements.txt"))
+        with open(requirements_path, "w") as req_file:
+            for _dep in self.dependencies:
+                req_file.write(f"{_dep}\n")
+
+    def add_dependency(self, dependency: str) -> None:
+        if self.initialized:
```
alexpearce, line 38: This variable is never set.

I guess this and other things like it would be caught in tests. Could you please add some tests of this class?

pipeline/objects/environment/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+                been initialized."
+            )
+        self.dependencies.append(dependency)
+
+    def merge_with_environment(self, env: "Environment") -> None:
+        if not isinstance(env, Environment):
+            raise Exception("Can only merge with another environment")
+
+        self.dependencies.extend(env.dependencies)
+        self.extra_index_urls.extend(env.extra_index_urls)
```
alexpearce, line 50: This list doesn't seem to be used.

pipeline/objects/environment/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+                "Cannot add dependency after the environment has \
+                been initialized."
+            )
+        self.dependencies.append(dependency)
+
+    def merge_with_environment(self, env: "Environment") -> None:
+        if not isinstance(env, Environment):
+            raise Exception("Can only merge with another environment")
+
+        self.dependencies.extend(env.dependencies)
```
alexpearce, line 49: What happens for conflicting dependencies? e.g. if I merge an environment with `black==20.0.0` with an environment with `black==21.0.0`.

examples/docker/gpt_neo_docker.py
```diff
@@ -0,0 +1,52 @@
+from pipeline import Pipeline, Variable, docker, pipeline_function, pipeline_model
```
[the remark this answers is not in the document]
paulcjh: For sure, will add

examples/docker/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+## API architecture
+
+The auto-docker feature from the Pipeline library extends the base docker image found here: [pipeline-docker](https://github.com/mystic-ai/pipeline-docker).
+
+There are three main Docker images that will be added into the final docker compose that are required to run the API:
+1. `pipeline-docker` - The main image containing the API.
+2. `postgresql` - This is used for traffic logging, and will be extended to include auth.
+3. `redis` - For session data.
```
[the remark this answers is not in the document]
paulcjh: Yeah, I'll do that in a follow up PR as it requires a lot of changes in the pipeline-docker repo. The purpose of this doc is to reflect the "as is" state

examples/docker/README.md
```diff
@@ -0,0 +1,123 @@
...
+from pipeline import docker
+
+docker.create_pipeline_api([gptneo_pl], environment=pl_env)
+```
+
+Running this final command will generate 4 files in your local directory:
+1. `docker-compose.yml` - A docker compose file containing environment variables, build info etc.
+2. `Dockerfile` - The Dockerfile that is used to create the final container
+3. `gptneo.graph` - The serialised graph createed by the Pipeline context manager
+4. `requirements.txt` - The python dependencies as defined above (`["transformers==4.24.0","torch==1.13.0"]`)
```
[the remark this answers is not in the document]
paulcjh: It will overwrite. This optionality/error/warning will be present in a follow up PR

pipeline/objects/environment/__init__.py
paulcjh: This should be removed by me, it was left when checking out another branch.
Agreed on the necessity for tests though

pipeline/objects/environment/__init__.py
paulcjh: To be removed, pulled from another branch. Will fix

pipeline/objects/environment/__init__.py
paulcjh: At the moment there's no conflict checks. I believe the way that pip install from a `requirements.txt` the last most entry will be the one present in the final system.
This needs a lot more work, I think I'll just remove the merging for the moment actually.

paulcjh: I'll remove the merging, it's not necessary right now.

examples/docker/gpt_neo_docker.py
```diff
@@ -0,0 +1,76 @@
...
+
+you can test the running images:
+
+
+curl --request POST \
+  --url http://localhost:5010/v2/run \
+  --header 'Content-Type: application/json' \
+  --data '{"pipeline_id":"gptneo", "data":"Hey there my name is"}'
+
+"""
```
alexpearce, line 22: Nice!

examples/docker/gpt_neo_docker.py
```diff
@@ -0,0 +1,76 @@
...
+        self.model = None
+        self.tokenizer = None
+
+    @pipeline_function
+    def predict(self, input_data: str) -> str:
+        if self.model is None:
+            from transformers import GPT2Tokenizer, GPTNeoForCausalLM
+
+            self.model = GPTNeoForCausalLM.from_pretrained("EleutherAI/gpt-neo-125M")
+            self.tokenizer = GPT2Tokenizer.from_pretrained("EleutherAI/gpt-neo-125M")
```
alexpearce, line 41: No preference from me, but why this style rather than the usual `load` function?

examples/docker/gpt_neo_docker.py
```diff
@@ -0,0 +1,76 @@
...
+            do_sample=True,
+            temperature=0.9,
+            max_length=100,
+        )
+        gen_text = self.tokenizer.batch_decode(gen_tokens)[0]
+        return gen_text
+
+
+with Pipeline("GPTNeo") as builder:
+    in_1 = Variable(str, is_input=True)
```
alexpearce, line 56: Just an aside, but seeing as this is a PR about cleanup: is `is_input` necessary here? Would a user ever create a output variable explicitly, e.g. I guess via `Variable(int, is_input=False)`? Could consider making `is_input=True` the default.

examples/docker/README.md
```diff
@@ -0,0 +1,123 @@
...
+curl --request POST \
+  --url http://localhost:5010/v2/run \
+  --header 'Content-Type: application/json' \
+  --data '{"pipeline_id":"gptneo", "data":"Hey there my name is"}'
+```
+
+## Not supported
+
+Currently the following Variables are not supported with the auto-docker feature:
+- PipelineFile
```
alexpearce, line 122: I guess this shouldn't be too painful in principle; the graph can figure out which inputs are pipeline files and expose this info, such that the Dockerfile can contain `COPY` statements acting on the source files.

examples/docker/gpt_neo_docker.py
paulcjh: They're mutually exclusive as the `is_output` field can be true even if the `is_input` field is true. 
These field are going to be removed in a future PR altogether and the context manager will set them instead of the user, it's so ugly at the moment. I'm happy with this example as is for the moment

examples/docker/README.md
paulcjh: Yeah for sure, totally agree. This is to work with the current 'as is' state, will add that in later

PR #167, opened 2022-11-29
Stable diffusion v2.0 auto-docker

paulcjh:
# Pull request outline

This PR adds in a new example for Stable diffusion v2.0 auto-docker, along with some necessary features to create the full docker image.

## Checklist:
- [x] Version bumped

~~- [ ] Docs updated~~
~~- [ ] Tests written~~
~~- [ ] Check for breaking changes in Resource~~

Added:
- `PipelineFile` support in auto-docker
- Support for GPU pass through in auto-docker
- Stable diffusion v2.0 auto-docker example
- Building the docker image from python

Changed:
- GPT Neo auto-docker now uses the new `/v2/runs` endpoint instead of the old `/v2/run`

Removed:
- Remove "hiring" tag to unused external endpoint

## Related issues:
_none_

___


examples/docker/stable_diffusion_2_0_docker.py
```diff
@@ -0,0 +1,273 @@
...
+    Pipeline,
+    PipelineFile,
+    Variable,
+    docker,
+    pipeline_function,
+    pipeline_model,
+)
+from pipeline.objects.environment import Environment
+
+# the below two functions don't seem to do their job
```
alexpearce, line 57: Worth keeping these lines then?

examples/docker/stable_diffusion_2_0_docker.py
```diff
@@ -0,0 +1,273 @@
...
+    os.environ["PL_GLOBAL_SEED"] = str(seed)
+    random.seed(seed)
+    np.random.seed(seed)
+    torch.manual_seed(seed)
+    torch.cuda.manual_seed_all(seed)
+    return seed
+
+
+def get_aspect_ratio(width: int, height: int) -> float:
+    return float(width / height)
```
alexpearce, line 93: Integer division always returns a float.

```suggestion
    return width / height
```

examples/docker/stable_diffusion_2_0_docker.py
```diff
@@ -0,0 +1,273 @@
...
+    def predict(
+        self, prompts: list[PromptShape], batch_kwargs: BatchKwargsShape
+    ) -> list[list[str]]:
+        print("🦁 run start")
+        print(prompts)
+        import base64
+        import random
+        from io import BytesIO
+
+        default_batch_kwargs = {
```
alexpearce, line 124: Because the inputs are Pydantic models, could you define the defaults in the model defintions instead?

pipeline/docker/__init__.py
```diff
@@ -110,5 +129,26 @@ def create_docker_compose(path, **environment_vars):
...
+            }
+        }
+
     with open(os.path.join(path, "docker-compose.yml"), "w") as docker_compose_file:
         docker_compose_file.write(yaml.dump(docker_yml_dict))
+
+
+def build():
+    subprocess.run("sudo docker compose pull".split(" "))
+    subprocess.run("sudo docker compose build".split(" "))
```
alexpearce, line 154: The default behaviour of `str.split` is to split on whitespace, so you could omit the argument here.

pipeline/docker/__init__.py
```diff
@@ -110,5 +129,26 @@ def create_docker_compose(path, **environment_vars):
             },
         },
     }
+
+    if gpu_index is not None:
```
rossgray, line 133: it looks like the value of `gpu_index` is never used?

examples/docker/stable_diffusion_2_0_docker.py
paulcjh: I'm going to update this completely in a follow up PR, not happy with the system at all.

pipeline/docker/__init__.py
paulcjh: fixed, good catch

PR #176, opened 2022-12-07
Improve error handling when server error returns non-JSON response

rossgray:
# Pull request outline

Currently when there is an error response from the API without a JSON body, the client receives a JSON decoding error rather than the actual error from the server.

For example, if the server returns a 503 error (which does not have a JSON body) the client will not see this, they will instead see something like `requests.exceptions.JSONDecodeError: Expecting value: line 1 column 1 (char 0)`

## Checklist:
~- [ ] Docs updated~
- [x] Tests written

- [x] Version bumped

Changed:
- Improved error handling when server error returns non-JSON response 

## Related issues:
_none_

### DIFF
--- a/pipeline/api/cloud.py
+++ b/pipeline/api/cloud.py
@@ -147,7 +147,11 @@ class PipelineCloud:
             isinstance(response, httpx.Response)
             and not response.status_code == httpx.codes.OK
         ):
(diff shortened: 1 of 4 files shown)


paulcjh (approved):
LGTM

PR #178, opened 2022-12-07
Preparation for supporting multiple file format uploads

rossgray:
# Pull request outline

This PR cleans up a few things related to file uploads and adds some new schemas to allow us to support uploading files in formats other than hex.

## Checklist:
~- [ ] Docs updated~
~- [ ] Tests written~
- [x] Version bumped

Added:
- `FileFormat` schema - will be used to support upload files in different formats (e.g. `binary`, possibly `gzip` in future)

Changed:
- Replaced `requests_toolbelt.multipart.encoder.MultipartEncoder` since this functionality is [supported by `requests` natively](https://requests.readthedocs.io/en/latest/user/quickstart/#post-a-multipart-encoded-file)
  - This also makes the code cleaner and easier to adapt to pass in additional form fields in the future

Removed:
- `requests-toolbelt` dependency (this is no longer needed since we can do `multipart/form-data` uploads using `requests`)
- unused `python_object_to_file_create` function
- `remote_path` argument for file uploads since this was not being used

## Related issues:
_none_

___


pipeline/api/cloud.py
```diff
@@ -335,50 +334,33 @@ def _post(self, endpoint, json_data):
...
         headers = {
             "Authorization": "Bearer %s" % self.token,
-            "Content-type": encoded_stream_data.content_type,
         }
         url = urllib.parse.urljoin(self.url, endpoint)
         response = requests.post(
-            url, headers=headers, data=encoded_stream_data, timeout=self.timeout
+            url,
+            headers=headers,
+            files={"file": (file.name, file, "application/octet-stream")},
```
alexpearce, line 362: Nice!

@isNeil looks like one could send multiple files this way 👀

alexpearce (commented):
Great clean-ups! Thank you.

isNeil:
Looks good to me, just waiting on clarification on whether the move away from file streaming affects anything.

alexpearce (commented):
Great work @rossgray and @isNeil! Just the CI that needs tending to then we should be ready to rock.

isNeil:
@alexpearce we've hit a bit of a roadblock with HTTPX, been unable to find a way to record uploading progress when hitting `upload_pipeline_file`.  We were using `CallbackIOWrapper(progress.update, data, "read")` from `tqdm` module for `requests.post`, which wraps the read model to update tqdm progress, but `httpx.post` only accepts `bytes`/`str` types.

pipeline/api/cloud.py
```diff
@@ -335,51 +353,36 @@ def _post(self, endpoint, json_data):
...
-                bar.n = monitor.bytes_read
-                bar.refresh()
-                if monitor.bytes_read == encoder_len:
-                    bar.close()
-
-        encoded_stream_data = encoder.MultipartEncoderMonitor(
-            e, callback=progress_callback if self.verbose else None
-        )
+            # If verbose then wrap our file object in a tqdm callback
+            file = CallbackIOWrapper(progress.update, file, "read")
```
alexpearce, line 372:
```suggestion
            file = CallbackIOWrapper(progress.update, file)
```

I guess this code path is lacking coverage.

tests/conftest.py
```diff
@@ -39,128 +39,143 @@ def test_with_decorator():
...
+def url():
+    return "http://127.0.0.1:8080"
+
+
+@pytest.fixture(scope="session")
+def httpserver_listen_address():
+    # Define the listen address for pytest-httpserver at the session scope so
+    # we can use it in defining environment variables before app creation
+    # https://pytest-httpserver.readthedocs.io/en/latest/howto.html#fixture
+    return ("127.0.0.1", 8080)
```
alexpearce, line 52: I guess these two values should be in sync, so how about we make one dependent on the other.

```suggestion
@pytest.fixture(scope="session")
def httpserver_listen_address():
    # Define the listen address for pytest-httpserver at the session scope so
    # we can use it in defining environment variables before app creation
    # https://pytest-httpserver.readthedocs.io/en/latest/howto.html#fixture
    return ("127.0.0.1", 8080)


@pytest.fixture(scope="session")
def url(httpserver_listen_address):
    host, port = httpserver_listen_address
    return f"http://{host}:{port}"
```

tests/test_pipeline_cloud.py
```diff
@@ -84,8 +82,8 @@ def test_cloud_upload_pipeline_file(
     )
 
 
-@pytest.mark.usefixtures("api_response")
-def test_cloud_get_raise_for_status_when_non_json_error(url, token):
+# TODO check if actually fixed
```
alexpearce, line 85: Has this been resolved?

pipeline/api/cloud.py
```diff
@@ -55,6 +54,24 @@
...
+    """
+
+    def __init__(self, callback: Callable, initial_bytes: bytes):
+        self._callback = callback
+        super().__init__(initial_bytes)
+
+    def read(self, size=-1) -> bytes:
+        data = super().read(size)
+        self._callback(len(data))
+        return data
```
rossgray, line 72: not sure if worth moving this into the `util` folder?

pipeline/api/cloud.py
rossgray: @alexpearce Rather confusingly, we are currently using 2 of these callback wrapper classes: `CallbackIOWrapper` is the one provided by the `tqdm` library, and `CallbackBytesIO` is the one I created (needed to work with `httpx`).

So this code was working fine before but still works now (`"read"` is the default for this argument).

If we want to just use my class for this for simplicity then we could do (we would just need to allow it to accept a `BytesIO` as an input in addition to `bytes`).

Hope that makes sense?

pipeline/api/cloud.py
isNeil: moved

poetryold.lock
```diff
@@ -0,0 +1,921 @@
+[[package]]
```
alexpearce, line 1: This can be deleted.

alexpearce (changes requested):

poetryold.lock
isNeil: ah yeah forgot about that

pipeline/api/cloud.py
alexpearce: Ahh, I missed that there were two classes with the same name!

No strong preference between consolidation or using the two separately; as you like.

pipeline/schemas/file.py
```diff
@@ -13,6 +13,7 @@ class FileFormat(str, enum.Enum):
 
 class FileBase(BaseModel):
     name: str
+    # TODO(RG): hex is currently the default but this will soon change
```
rossgray, line 16: think this line was left over from a previous commit of mine (you can probably remove it)

alexpearce (dismissed):
Nice! I'm happy with this if you are @rossgray.

pyproject.toml
```diff
@@ -22,7 +22,6 @@ pydantic = "^1.8.2"
 requests = "^2.26.0"
```
rossgray, line 22: can `requests` also be removed now?

pipeline/api/cloud.py
```diff
@@ -12,7 +12,6 @@
 import httpx
 import requests
```
rossgray, line 13: I think `requests` is still referenced in `_get_raise_for_status(...)` so if this is updated it can be removed entirely

pyproject.toml
```diff
@@ -33,14 +31,14 @@ packaging = "^22.0"
...
 [tool.poetry.dev-dependencies]
 pydantic = "^1.8.2"
-requests = "^2.26.0"
 pyhumps = "^3.0.2"
 pytest = "^6.2.5"
 black = "^22.10.0"
 pre-commit = "^2.16.0"
 flake8 = "^5.0.4"
 isort = "^5.10.1"
 responses = "^0.20.0"
```
rossgray, line 40: can `responses` also be removed?

isNeil:
cool, thanks for the pointers, think I've cleaned it up now

rossgray:
Thanks for taking over this PR @isNeil. Nice refactor of all the test code. I can't approve this since I'm the original author but happy for you or Alex to approve

alexpearce (dismissed):

isNeil (approved):
bumped v since Alex's review

PR #179, opened 2022-12-09
Add file format schemas

rossgray:
# Pull request outline

Updating schemas in order to support pipeline file uploads in formats other than hex. 

## Checklist:
~- [ ] Docs updated~
~- [ ] Tests written~
- [x] Version bumped

Added:
- `FileFormat` schema

Changed:
- File-related schemas to support passing in the file format

## Related issues:
_none_

### DIFF
--- a/pipeline/schemas/pipeline_file.py
+++ b/pipeline/schemas/pipeline_file.py
@@ -1,11 +1,13 @@
-from typing import List
+from typing import List, Optional
 
 from .base import BaseModel
-from .file import FileGet
+from .file import FileFormat, FileGet
 
 
 class PipelineFileDirectUploadInitCreate(BaseModel):
     file_size: int
+    # hex is the default purely for backwards-compatability
+    file_format: FileFormat = FileFormat.hex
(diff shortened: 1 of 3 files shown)


pipeline/schemas/pipeline_file.py
```diff
@@ -42,4 +44,7 @@ class PipelineFileDirectUploadFinaliseCreate(BaseModel):
 class PipelineFileGet(BaseModel):
     id: str
     name: str
-    hex_file: FileGet
+    # This is a legacy field that is replaced by 'file'.
+    # Can be deprecated once no clients are using it.
```
alexpearce, line 48: I suppose we can consider it deprecated immediately, and then _remove_ it once it's no longer used.

Pydantic has no explicit support for deprecated fields from what I can see, but one could use a `Field(None, deprecated=True)` convention if desired. Our versioning is so informal today that it probably doesn't mean much to add this though.

alexpearce (approved):
LGTM!z

PR #190, opened 2022-12-21
Use multipart upload API for file-backed resources

isNeil:
Switch to the new multipart upload strategy for creating API resources backed by files. This affects the creation of Data, Function, Model, and Pipeline.

The previous strategy, soon to be deprecated in the API, was to upload each backing files to the `POST /v2/files/` endpoint and then use the resulting File IDs in a separate request to create the resource, e.g. a Function.

The new strategy submits a single [multipart request](https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html) which consists of one part for the usual JSON payload, minus any File IDs, and one part for each backing file.

The new strategy has a few of advantages:

1. The client only makes a single request, reducing total execution time.
2. Backing files are now uploaded in their binary form, rather than being hex-encoded, reducing execution time by removing the hex-encoded step and reducing bandwidth usage and the binary form is a factor 2 smaller.
3. Resource creation is now atomic with the creation of backing files, elimanating the possibility of creating 'dangling' Files (Files not referenced by a resource).

## Checklist:
- [x] ~~Docs updated~~ not required as no user-facing changes.
- [x] Tests written
- [x] Version bumped to v0.3.5.

Added:
- Helpers for loading and dumping which are compatible between both hex-encoded and binary payloads.

Changed:
- Use multipart requests for creating Data, Function, Model, and Pipeline resources.
- `file` and `file_id` fields of `*Create` schemas are now marked as deprecated. API clients should instead provide files as part of a single multipart request rather than including a `FileGet` or File ID.

Removed:
- Unused `FunctionIOCreate` schema.

## Related issues:

Towards BE-207.

### DIFF
--- a/pipeline/api/cloud.py
+++ b/pipeline/api/cloud.py
@@ -7,7 +7,18 @@ import os
 import sys
 import uuid
 from http import HTTPStatus
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Type, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    BinaryIO,
+    Dict,
+    List,
+    Optional,
+    Set,
+    Tuple,
+    Type,
+    Union,
+)
 
 import dill
 import httpx
@@ -23,14 +34,14 @@ from pipeline.objects.variable import PipelineFile
 from pipeline.schemas.base import BaseModel
 from pipeline.schemas.compute_requirements import ComputeRequirements
 from pipeline.schemas.data import DataGet
-from pipeline.schemas.file import FileCreate, FileFormat, FileGet
+from pipeline.schemas.file import FileFormat, FileGet
 from pipeline.schemas.function import FunctionCreate, FunctionGet
 from pipeline.schemas.model import ModelCreate, ModelGet
 from pipeline.schemas.pipeline import (
     PipelineCreate,
     PipelineFileVariableGet,
     PipelineGet,
-    PipelineVariableGet,
+    PipelineVariableCreate,
 )
 from pipeline.schemas.pipeline_file import (
     MultipartUploadMetadata,
@@ -44,10 +55,9 @@ from pipeline.schemas.pipeline_file import (
 from pipeline.schemas.run import RunCreate, RunGet
 from pipeline.util import (
     CallbackBytesIO,
-    generate_id,
-    hex_to_python_object,
+    dump_object,
+    load_object,
     package_version,
-    python_object_to_hex,
     python_object_to_name,
 )
 from pipeline.util.logging import PIPELINE_FILE_STR, PIPELINE_STR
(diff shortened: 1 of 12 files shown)


pipeline/api/cloud.py
```diff
@@ -169,37 +198,20 @@ def _get_raise_for_status(response: httpx.Response) -> None:
...
-            )
-        else:
-            return self._post_file("/v2/files/", file_or_path)
-
-    def upload_data(self, file_or_path) -> DataGet:
-        uploaded_file = self.upload_file(file_or_path)
-        uploaded_data = self._post("/v2/data", uploaded_file.dict())
+    def upload_data(self, obj: Any) -> DataGet:
+        """Upload the object as a pickle-encoded file to the API."""
+        files = dict(data=_as_upload_file(_as_binary_file(obj)))
```
rossgray, line 203: it seems like we always call `_as_upload_file` with `_as_binary_file` so I wonder if it's worth calling `_as_binary_file` inside `_as_upload_file` just to clean this up a bit and make the caller's code a bit simpler?

pipeline/api/cloud.py
```diff
@@ -55,6 +55,35 @@
...
+
+    Returns a structure suitable for passing to the `files` parameter of
+    httpx.request.
+    """
+    if name is None:
+        name = str(uuid.uuid4())
+    return (name, object, BINARY_MIME_TYPE)
+
+
+def _ensure_binary_file(file_or_path: Union[BinaryIO, str, pathlib.Path]) -> BinaryIO:
```
rossgray, line 72: I can only see this being used in one place (by `_as_binary_file`) so just checking it is actually needed?

pipeline/schemas/pipeline.py
```diff
@@ -24,6 +24,39 @@ class PipelineFileVariableGet(BaseModel):
     file: FileGet
 
 
+class PipelineVariableCreate(BaseModel):
+    local_id: str
+    name: Optional[str]
+
+    type_file: Optional[FileGet] = Field(
+        default=None,
+        deprecated=True,
```
rossgray, line 33: ah nice!

rossgray (commented):
Just a couple of small comments. I presume the changes to top would need to made at the same time to ensure these updated API calls work?

pipeline/api/cloud.py
alexpearce: Good catch! It isn't needed. The `instance(file_or_path, str)` check always failed because the argument is always a `io.BytesIO`.

💣

pipeline/objects/function.py
```diff
@@ -53,11 +53,16 @@ def __init__(
 
     @classmethod
     def from_schema(cls, schema: FunctionGet):
-        unpickled_data = hex_to_python_object(schema.hex_file.data)
+        unpickled_data = load_object(schema.hex_file.data)
```
rossgray, line 56: I presume we can't get away from the `hex_file` name due to backwards-compatability?

rossgray (approved):
Changes look good to me

pipeline/objects/function.py
alexpearce: Indeed 😞

PR #205, opened 2023-01-18
Pipeline tag regex update

plutopulp:
Changed:
- Fix regex mismatch for pipeline tag names between pipeline `CLI` and `pipeline-top`


alexpearce (approved):
Good catch!

PR #207, opened 2023-01-19
Set pipeline-ai version in user agent

alexpearce:
Setting the [user agent header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent) to `pipeline-ai/{version}` will allow us to track usage per package version. This will be helpful in understanding when deprecated APIs can be removed.

## Checklist:
- [ ] ~~Docs updated~~
- [x] Tests written
- [ ] ~~Check for breaking changes in Resource~~
- [x] Version bumped

Added:
- N/A.

Changed:
- `PipelineCloud` now sends a `User-Agent` header of the form `pipeline-ai/{version}` when talking to the API.
- Refactored HTTP interactions in `PipelineCloud` to use a single HTTP client instance. Parameters common to all API requests, such as the `Authorization` header and the timeout, are now defined and used in a single place, making modifications easier.

Removed:
- N/A.

## Related issues:
_none_

### DIFF
--- a/pipeline/api/cloud.py
+++ b/pipeline/api/cloud.py
@@ -5,7 +5,6 @@ import io
 import json
 import os
 import sys
-import urllib.parse
 import uuid
 from http import HTTPStatus
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Type, Union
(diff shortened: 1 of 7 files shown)


pipeline/api/cloud.py
```diff
@@ -308,64 +308,34 @@ def upload_pipeline_file(
         )
 
     def _get(self, endpoint: str, params: Dict[str, Any] = None):
-        headers = {
-            "Authorization": "Bearer %s" % self.token,
-        }
-
-        url = urllib.parse.urljoin(self.url, endpoint)
-        response = httpx.get(url, headers=headers, params=params, timeout=self.timeout)
+        self.raise_for_invalid_token()
```
alexpearce, line 311: @paulcjh I noticied that `_get` was unique in not having `raise_for_invalid_token`. I added it in case it was an oversight, but perhaps it's missing intentionally?

paulcjh: No, nice spot!

alexpearce:
Thanks for the review @paulcjh! Just fixed the isort check, this should be good to go now.

PR #209, opened 2023-01-26
Fix PipelineFile upload

rossgray:
# Pull request outline

This PR fixes a regression when the http client was updated.

## Checklist:
- [x] Version bumped

Changed:
- PipelineFile uploads fixed


alexpearce (approved):
Thanks very much for the speedy fix! 🙇

PR #211, opened 2023-01-31
Send Environment ID during Pipeline creation.

alexpearce:
The `PipelineCreate` schema now includes an _optional_ `environment_id` field, which defines the execution environment a pipeline will be run within on the compute resource. It must be optional to allow the API to support older clients.

The set of mystic-provided 'default' environments is hard-coded into the client as the new `pipeline.api.environments` module such that scripts can programatically reference a default if they wish.

Because environments are a Pipeline-Cloud-specific feature (local execution will use whatever environment the user has installed), I've gone for adding an `environment` argument to `PipelineCloud.upload_pipeline` rather than the `Pipeline` constructor. The `environment` argument is currently optional, and when it is `None` the client falls back to a fixed default environment. The name `environment` hints at the possibility of us supporting non-ID values in the future (e.g. names, or some other Python object representation).

## Checklist:
- [ ] ~~Docs updated~~ @paulcjh I've held off doing this as the specifics of what the current default environment contains are TBD, and can anyhow be updated at a later date. I don't think the broader docs of 'custom environment support' are within the scope of this PR.
- [x] Tests written
- [ ] ~~Check for breaking changes in Resource~~
- [x] Version bumped

### DIFF
--- a/pipeline/api/cloud.py
+++ b/pipeline/api/cloud.py
@@ -16,6 +16,7 @@ from tqdm import tqdm
 from tqdm.utils import CallbackIOWrapper
 
 from pipeline import configuration
+from pipeline.api.environments import PipelineCloudEnvironment, resolve_environment_id
 from pipeline.exceptions.InvalidSchema import InvalidSchema
 from pipeline.exceptions.MissingActiveToken import MissingActiveToken
 from pipeline.objects.variable import PipelineFile
@@ -55,6 +56,7 @@ if TYPE_CHECKING:
     from pipeline.objects import Function, Graph, Model
 
 FILE_CHUNK_SIZE = 200 * 1024 * 1024  # 200 MiB
+EnvironmentObjectOrID = Union[PipelineCloudEnvironment, str]
 
 
 class PipelineCloud:
@@ -421,6 +423,7 @@ class PipelineCloud:
         public: bool = False,
         description: str = "",
         tags: Set[str] = None,
+        environment: Optional[EnvironmentObjectOrID] = None,
     ) -> PipelineGet:
         """
         Upload a Pipeline to the Cloud.
@@ -432,6 +435,10 @@ class PipelineCloud:
                         Defaults to False.
                     description (str): Description of the Pipeline.
                     tags (Set[str]): Set of tags for the pipeline. Eg: {"BERT", "NLP"}
+                    environment (Optional[Union[PipelineCloudEnvironment, str]]):
+                        Identifier of the execution environment the pipeline
+                        should run within. If None (the default) a Mystic-
+                        provided default environment will be chosen.
 
             Returns:
                     pipeline (PipelineGet): Object representing uploaded pipeline.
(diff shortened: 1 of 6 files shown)


paulcjh (commented):
Thanks for this. I do have a bit of hesitation around this implementation for two reasons:
1. There's already an `Environment` object present in the library, and here another one is defined just in the cloud context with a hard coded ID. Seems confusing, and they're not compatible.
2. Is there a reason why the default ID is stored here instead of assigned on top, and then we can make the `environment_id` field optional on the `PipelineCreate` schema?

pipeline/api/cloud.py
```diff
@@ -421,6 +422,7 @@ def upload_pipeline(
         public: bool = False,
         description: str = "",
         tags: Set[str] = None,
+        environment: Optional[Union[PipelineCloudEnvironment, str]] = None,
```
paulcjh, line 425: The name here is a bit confusing, perhaps `environment_object_or_id` or something to that effect might be better.

pipeline/api/cloud.py
```diff
@@ -421,6 +422,7 @@ def upload_pipeline(
...
     ) -> PipelineGet:
```
paulcjh, line 426: Doc string needs updating with the new variable

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "pipeline-ai"
-version = "0.3.11"
+version = "0.3.12"
```
paulcjh, line 3: As this feature is included under the v0.4.0 spec should this be reflecting that?

paulcjh (changes requested):

alexpearce:
Thanks for the review! On your points:

> There's already an Environment object present in the library

I thought that was only used for 'local' environment creation, and that wasn't something we wanted to continue with.

TBH the container I made really only needs to be the IDs. I only made the container to store the name, which I figured would be useful for browsing envs on the dashboard (where the name is the primary identifier).

> Is there a reason why the default ID is stored here instead of assigned on top

Mostly because I didn't like the idea of top deciding what the default was. By keeping on the client it's transparent to the user what environment they will get (the one in the Python module), whereas when it's on pipeline-top what they get may change from one day to the next.

pipeline/api/cloud.py
alexpearce: I try to avoid putting the types into the name as that information is already encoded in the type and/or the docstring. But it's not a particularly strong preference so am happy to change if you'd like.

pyproject.toml
alexpearce: Perhaps yes, although to me the real feature is _creating_ environments, which this PR does not expose.

As this PR includes a schema change, if we want for the 0.4.0 release we won't be able to test the 'read' aspect of the API (i.e. this default environment behaviour) until everything is merged.

pipeline/api/cloud.py
rossgray: another option that I've seen used is to create a type alias, eg something like:

```python
EnvironmentObjectOrId = Union[PipelineCloudEnvironment, str]

...
environment: Optional[EnvironmentObjectOrId] = None,
...
```

alexpearce: Yeah, that would make the annotation more readable.

With Python 3.10 it could become even nicer: `PipelineCloudEnvironment | str | None`.

alexpearce: I've factored out the type.

paulcjh (approved):

PR #212, opened 2023-02-01
add base environment schema + attach to pipeline detailed

plutopulp:
# Pull request outline

Added:
- A Base Environment Schema

Changed:
- Added Base environment to pipeline get detailed schema

## Related issues:
We want to attach environment information to pipeline responses. We could just attach the environment id. However, when a user deletes an environment, it would be useful for clients to be aware the environment has been deleted, avoiding them having to make an additional API and getting a `404 not found`. It is probably not a field we want to return in all EnvironmentGet responses, as the user will always see it as `None` if env exists, and then a 404 when it gets soft-deleted.

___

### DIFF
--- a/pipeline/schemas/environment.py
+++ b/pipeline/schemas/environment.py
@@ -1,3 +1,4 @@
+from datetime import datetime
 from typing import List, Optional
 
 from pipeline.schemas.base import BaseModel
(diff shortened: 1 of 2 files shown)


alexpearce (approved):
Seems reasonable to me 👍

PR #218, opened 2023-03-07
Plutopulp/default and public environments

plutopulp:
# Pull request outline

In this PR we we add to the environments CLI, where we:
- Refactor existing logic for CRUD by_name on environments by resolving/fetching the environment right off the bat, by `name` or `id` and then use the ID of the resulting object to make further actions on that environment.
- Add logic for listing publicly available environments, e.g. `pipeline environments list --public`
- Add logic for retrieving the default environment, e.g. `pipeline environments get --default`

Note that the default environment configuration `(id, name)` is always expected to be defined in this client, i.e. `pipeline-ai` and not in the `API` itself, even if the environment object associated with this configuration, is defined in the API. This lets the client choose what the default environment should be. Just something to note.
Further, the way we have implemented getting the default environment isn't the prettiest, where we conditionally add the required parser argument `name_or_id` depending on whether the `--default` flag has been set. This makes the UX of the CLI potentially nicer, but the following for instance, would keep things less convoluted: `pipeline environments get-default`. 

## Checklist:
- [ ] Docs updated
- [x] Tests written
- [x] Version bumped

___

### DIFF
--- a/pipeline/console/__init__.py
+++ b/pipeline/console/__init__.py
@@ -289,10 +289,11 @@ def main(args: Optional[List[str]] = None) -> int:
     )
     # Positional arguments are required. We don't want to have to include name_or_id
     # when the default flag is set
-    "--default" not in sys.argv and environments_get_parser.add_argument(
-        "name_or_id",
-        help="The environment name or id to get",
-    )
+    if "--default" not in sys.argv:
+        environments_get_parser.add_argument(
+            "name_or_id",
+            help="The environment name or id to get",
+        )
 
     environments_get_parser.add_argument(
         "-n",
--- a/pipeline/console/__init__.py
+++ b/pipeline/console/__init__.py
@@ -282,9 +282,16 @@ def main(args: Optional[List[str]] = None) -> int:
         "get",
         help="Get environment information",
     )
-
     environments_get_parser.add_argument(
-        "name_or_id", help="The environment name or id to get"
+        "--default",
+        help="Get the default environment",
+        action="store_true",
+    )
+    # Positional arguments are required. We don't want to have to include name_or_id
+    # when the default flag is set
+    "--default" not in sys.argv and environments_get_parser.add_argument(
+        "name_or_id",
+        help="The environment name or id to get",
     )
 
     environments_get_parser.add_argument(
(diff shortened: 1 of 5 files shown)


pipeline/console/__init__.py
```diff
@@ -282,9 +282,16 @@ def main(args: Optional[List[str]] = None) -> int:
...
-
     environments_get_parser.add_argument(
-        "name_or_id", help="The environment name or id to get"
+        "--default",
+        help="Get the default environment",
+        action="store_true",
+    )
+    # Positional arguments are required. We don't want to have to include name_or_id
+    # when the default flag is set
+    "--default" not in sys.argv and environments_get_parser.add_argument(
```
alexpearce, line 292: I'd prefer `if "--default" not in sys.argv:` but won't insist.

pipeline/console/environments.py
```diff
@@ -18,37 +19,46 @@
...
     remote_service = PipelineCloud(verbose=False)
     remote_service.authenticate()
 
-    url = (
-        f"/v2/environments/by-name/{name_or_id}"
-        if by_name
-        else f"/v2/environments/{name_or_id}"
-    )
-
+    url = get_url(name_or_id, by_name, default)
```
alexpearce, line 41: Ahh, if only Python `if`s were expressions…

pipeline/console/environments.py
```diff
@@ -84,87 +94,75 @@ def _create_environment(name: str, packages: List[str]) -> EnvironmentGet:
...
-        f"/v2/environments/{environment_information.id}",
+        f"/v2/environments/{id}",
         update_schema.dict(),
     )
     return EnvironmentGet.parse_obj(response)
 
 
 def _add_packages_to_environment(
-    name_or_id: str,
+    environment: EnvironmentGet,
```
alexpearce, line 132: I like the way you pass in this environment object!

alexpearce (dismissed):
Nice refactor and nice tests to go with it. Thank you!

pipeline/console/__init__.py
plutopulp: Yeah I hear you ;) Updated. Thx

pipeline/console/environments.py
plutopulp: Yeah, made sense since we're fetching it off the bat and needed other non-id properties in the method.

alexpearce (approved):
🚢

PR #222, opened 2023-05-02
Paul/pcore

paulcjh:
# Pull request outline

## Checklist:
- [ ] Docs updated
- [ ] Tests written
- [ ] Version bumped

Added:
- A list of new things

Changed:
- A list of things changed

Removed:
- A list of things removed

## Related issues:
_none_

___


pipeline/v3/http.py
```diff
@@ -0,0 +1,112 @@
...
+ACTIVE_IP = (
+    active_remote.url
+    if (active_remote := current_configuration.active_remote) is not None
+    else None
+)
+
+_client = httpx.Client(
+    base_url=ACTIVE_IP,
+    headers={
+        "Authorization": f"Bearer {current_configuration.active_remote.token}",
```
rossgray, line 21: would be nice to be able to use env vars for this stuff too, like for v2 (especially if people are using pipeline from within their own applications)

PR #246, opened 2023-08-03
Add support for returning errors

rossgray:
# Pull request outline

## Checklist:
- [ ] Docs updated
- [ ] Tests written
- [ ] Version bumped

Added:
- A list of new things

Changed:
- A list of things changed

Removed:
- A list of things removed

## Related issues:
_none_

___


pipeline/cloud/pipelines.py
```diff
@@ -241,7 +243,7 @@ def map_pipeline_mp(array: list, graph_id: str, *, pool_size=8):
 def stream_pipeline(
     pipeline_id_or_pointer: str,
     *data,
-) -> t.Iterator[RunOutput]:
+) -> t.Iterator[t.Any]:
```
paulcjh, line 246:
```suggestion
) -> t.Iterator[t.Union[RunOutput, str]]:
```

pipeline/cloud/pipelines.py
```diff
@@ -253,7 +255,12 @@ def stream_pipeline(
     ) as generator:
         for item in generator.iter_text():
             if item:
-                yield RunOutput.parse_raw(item)
+                try:
+                    output = RunOutput.parse_raw(item)
+                    yield output.value
```
paulcjh, line 260:
```suggestion
                    yield output
```

pipeline/cloud/pipelines.py
```diff
@@ -253,7 +255,12 @@ def stream_pipeline(
...
+                except ValidationError:
+                    output = RunError.parse_raw(item)
+                    yield output.traceback or output.exception
```
paulcjh, line 263:
```suggestion
                    yield output
```

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "pipeline-ai"
-version = "1.0.0b1"
+version = "1.0.0b2"
```
paulcjh, line 3:
```suggestion
version = "1.0.0b3"
```

paulcjh (changes requested):
Nice, few changes needed then good to go.

pipeline/cloud/pipelines.py
rossgray: If I make the changes below the return type could either be a `RunOutput` or a `RunError`. The means that a user calling `stream_pipeline` will need to do something like the following:

```python
for output in stream_pipeline(pointer, *inputs):
    if isinstance(output, RunOutput):
        print(output.value)
    elif isinstance(output, RunError):
        print(output.traceback or output.exception)
```

which adds more work for the user and seems a little messy. What do you think?

paulcjh: I actually think a better thing to do would be to raise an exception?

rossgray: yeah good idea!

paulcjh (approved):
LGTM, nice work @rossgray

PR #275, opened 2023-09-04
Switch protocol correctly

MPCherry:
# Pull request outline

## Checklist:
- [ ] Docs updated
- [ ] Tests written
- [ ] Version bumped

Added:
- A list of new things

Changed:
- A list of things changed

Removed:
- A list of things removed

## Related issues:
_none_

___


paulcjh (changes requested):
Just need to VB then good to go

PR #313, opened 2023-09-19
Logging in local time

MPCherry:
https://linear.app/mystic-ai/issue/BE-780/change-timestamp-from-python-client-to-users-time-zone


plutopulp (approved):
LGTM :)

PR #316, opened 2023-09-26
Wrap run pipeline in retry

MPCherry:


rossgray (commented):
Looks good. The only question I have is around async runs since I'm not quite sure if it will work correctly in this case? (Typically with async runs you get a response back straight away before the run has reached a terminal state). Therefore I don't know if it makes sense to add retries for async runs or not? I suppose the question is similar for debug runs. Probably need @paulcjh 's input here

MPCherry:
> Looks good. The only question I have is around async runs since I'm not quite sure if it will work correctly in this case? (Typically with async runs you get a response back straight away before the run has reached a terminal state). Therefore I don't know if it makes sense to add retries for async runs or not? I suppose the question is similar for debug runs. Probably need @paulcjh 's input here

Ye this doesn't make sense at all for non-debug async runs, I think there's a risk here that customers will think retries will magically apply to their async runs, but we need some backend infrastructure for that. For debug runs this is working nicely as is in the PR.

```
Pipeline 16:36:44 - [INFO]: Trailing run logs (run_9048ab42cd864f1d8367be33bdbdf67c)
Pipeline 16:36:44 - [INFO]: Run state: unknown
Pipeline 16:37:09 - [INFO]: Run state: failed
Pipeline 16:37:09 - [INFO]: Run state was failed, sleeping for 5 seconds before retrying
Pipeline 16:37:15 - [INFO]: Trailing run logs (run_c24f1053288f42c18dffe3e68cd68391)
Pipeline 16:37:16 - [INFO]: Run state: unknown
Pipeline 16:37:41 - [INFO]: Run state: failed
Pipeline 16:37:41 - [INFO]: Run state was failed, sleeping for 5 seconds before retrying
```

rossgray:
> Ye this doesn't make sense at all for non-debug async runs, I think there's a risk here that customers will think retries will magically apply to their async runs, but we need some backend infrastructure for that.

Maybe it's worth checking for this case (where `async_run is True and retry is True`) and either returning an error or logging a warning?

rossgray (approved):
LGTM

rossgray (approved):

paulcjh:
Yup I like it, LGTM

PR #320, opened 2023-10-09
Pipeline accelerator validation methods

MPCherry:


plutopulp (approved):
LGTM! I like the `max_memory_mb` method

PR #321, opened 2023-10-09
Add no_resources_available terminal state

MPCherry:
Snuck in a change to the VRAM as well.

And generalized the terminal states a tad, so we can simplify some things in other places.


rossgray (approved):
Nice, thanks for the cleanup too!

rossgray (approved):

PR #324, opened 2023-10-12
add run inut file url

plutopulp:
# Pull request outline
Add optional external file URL  to input files 

## Checklist:
- [x] Version bumped

### DIFF
--- a/pipeline/cloud/schemas/runs.py
+++ b/pipeline/cloud/schemas/runs.py
@@ -222,6 +222,7 @@ class RunInput(BaseModel):
 
     file_name: t.Optional[str]
     file_path: t.Optional[str]
+    # The file URL is only populated when this schema is returned by the API, the user should never populate it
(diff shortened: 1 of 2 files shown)


pipeline/cloud/schemas/runs.py
```diff
@@ -222,6 +222,7 @@ class RunInput(BaseModel):
 
     file_name: t.Optional[str]
     file_path: t.Optional[str]
+    file_url: t.Optional[str]
```
paulcjh, line 225:
```suggestion
    # The file URL is only populated when this schema is returned by the API, the user should never populate it
    file_url: t.Optional[str]
```

paulcjh: Weird change, but don't want to loose track

paulcjh (changes requested):

paulcjh (approved):
LGTM

PR #330, opened 2023-10-25
pipeline metadata field alias

plutopulp:
# Pull request outline

Since Pydantic filters out field names which begin with an underscore, and we want to return a pipelines' metadata as `_metadata: {...}`, we get around this by creating an alias for the field. 

## Checklist:
- [x] Version bumped

___


MPCherry (approved):
LGTM

PR #336, opened 2023-11-06
Plutopulp/pipeline build from anywhere

plutopulp:
This PR allows for building a pipeline from any directory. Note that currently, no published package with new container stuff on pypi, so to override the pipeline-ai (user defined) package for dev, bind as volume, e.g. `docker run -p 14300:14300 -v ./pipeline:/usr/local/lib/python3.10/site-packages/pipeline plutopulp/add-lol`


paulcjh:
just going to eat lunch then I'll have a gander

examples/docker/from_anywhere/pipeline.yaml
```diff
@@ -0,0 +1,15 @@
...
+  container_commands:
+    - "apt update -y"
+    - "apt install -y git"
+    # - "apt-get update && \
+    #   apt-get -y install gcc mono-mcs && \
+    #   rm -rf /var/lib/apt/lists/*"
+  python:
+    python_version: "3.10"
+    python_requirements:
+      - "pipeline-ai==1.0.26"
```
plutopulp, line 11: May want to raise an error when the user forgets to provide a pipeline-ai version from now on?

rossgray: another option would be to use the version the user is currently using

examples/docker/from_anywhere/pipeline.yaml
```diff
@@ -0,0 +1,15 @@
+runtime:
+  container_commands:
+    - "apt update -y"
+    - "apt install -y git"
+    # - "apt-get update && \
+    #   apt-get -y install gcc mono-mcs && \
+    #   rm -rf /var/lib/apt/lists/*"
+  python:
+    python_version: "3.10"
```
rossgray, line 9: not related to this PR, but just commenting on this while I'm here: wonder if we can drop the `python_` prefix since it's already in a `python` section, e.g. we could just have:

```
python:
  version: "3.10"
  requirements:
    - "blah==1.0"
```

examples/docker/from_anywhere/pipeline.yaml
plutopulp: Do you mean if they have an activated environment, use the pipeline-ai version from that?

rossgray: yeah, I mean they have to run `pipeline container build` to build the container, so must have pipeline-ai installed, so we should be able to pick up the version directly

PR #338, opened 2023-11-08
frontend: added outputs

adamleithp-mystic:
# Pull request outline

## Checklist:
- [ ] Docs updated
- [ ] Tests written
- [ ] Version bumped

Added:
- A list of new things

Changed:
- A list of things changed

Removed:
- A list of things removed

## Related issues:
_none_

___


pipeline/container/frontend/index.html
```diff
@@ -692,22 +919,10 @@
...
-              setDynamicFields(
-                generateDynamicFieldsFromIOVariables(data.input_variables)
-              );
-            })
-            .catch((err) => console.log(err));
-        }, []);
-        console.log(pipeline);
-        React.useEffect(() => {
-          setInputData(generateInputDataArrayFromDynamicFields(dynamicFields));
-        }, [dynamicFields]);
```
plutopulp, line 710: What handles the inputData updates now then? (When the form values change)

adamleithp-mystic: you're right!

adamleithp-mystic: re-added! good catch

pipeline/container/frontend/index.html
```diff
@@ -759,57 +972,265 @@
...
+                data.input_variables
+              );
+              // const dynamicFields =
+              //   generateDynamicFieldsFromIOVariables(mockInputIOVariables);
+
+              setDynamicFields(dynamicFields);
+
+              const inputData =
+                generateInputDataArrayFromDynamicFields(dynamicFields);
+              setInputData(inputData);
```
plutopulp, line 1168: I guess this is not needed anymore due the useEffect, e.g. inputData will be set twice on mount. But no biggie if you want to leave it and merge

PR #345, opened 2023-11-14
Plutopulp/frontend file inputs

plutopulp:


pipeline/container/frontend/index.html
```diff
@@ -1022,32 +1044,55 @@
...
-        });
+        }
 
         return array;
       }
 
       function App() {
         const [dynamicFields, setDynamicFields] = React.useState([]);
         const [isLoading, setIsLoading] = React.useState(false);
-        const [inputData, setInputData] = React.useState([]);
```
adamleithp-mystic, line 1050: I'm confused, how are we tracking this info now?

plutopulp: Yeah so have changed it so that inputData is only constructed when the [form gets submitted](https://github.com/mystic-ai/pipeline/blob/plutopulp/frontend-file-inputs/pipeline/container/frontend/index.html#L1300-L1316). The dynamicFields, like before, track changes in the form state, so we only need to construct the inputData from the dynamicFields once really. This also has the benefit that files will only be uploaded when submitting the form, rather than each time a file is selected.

adamleithp-mystic: I see! good optimisation. LGTM.

PR #358, opened 2023-12-12
V1.1.0 rebase

MPCherry:


rossgray (approved):
that's a big diff but changes look good from a brief look

PR #362, opened 2023-12-21
Plutopulp/handle raise for status

plutopulp:
In this PR we add a decorator for our API client request methods which handles catching all http status errors from the server and displaying them to the user. We also show the correlation-id of the request, which users can provide to the Mystic team for debugging purposes. Errors/Exceptions should no longer be handled on this client but instead directly on the server, and simply displayed to the user on the client.

### DIFF
--- a/pipeline/cloud/http.py
+++ b/pipeline/cloud/http.py
@@ -24,7 +24,7 @@ def get_response_error_dict(e: httpx.HTTPStatusError) -> t.Dict:
             detail = {"detail": detail}
     except (TypeError, KeyError):
         return {"message": "Something went wrong.", "response_json": e.response.json()}
-    return {**detail, "x-correlation-id": e.response.headers.get("x-correlation-id")}
+    return {**detail, "request_id": e.response.headers.get("x-correlation-id")}
 
 
 def handle_http_status_error(func):
--- a/pipeline/cloud/http.py
+++ b/pipeline/cloud/http.py
@@ -26,6 +26,7 @@ def get_response_error_dict(e: httpx.HTTPStatusError) -> t.Dict:
         return {"message": "Something went wrong.", "response_json": e.response.json()}
     return {**detail, "x-correlation-id": e.response.headers.get("x-correlation-id")}
 
+
 def handle_http_status_error(func):
     @functools.wraps(func)
     def wrapper(*args, **kwargs):
--- a/pipeline/cloud/http.py
+++ b/pipeline/cloud/http.py
@@ -31,7 +31,7 @@ def handle_http_status_error(func):
     @functools.wraps(func)
     def wrapper(*args, **kwargs):
         response = func(*args, **kwargs)
-        if kwargs.pop("raise_for_status", True):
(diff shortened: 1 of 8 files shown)


pipeline/cloud/http.py
```diff
@@ -10,12 +11,40 @@
...
+        response = func(*args, **kwargs)
+        if kwargs.pop("handle_error", True):
+            try:
+                response.raise_for_status()
+            except httpx.HTTPStatusError as e:
+                _print(
+                    get_response_error_dict(e),
+                    level="ERROR",
+                )
+                raise
```
rossgray, line 42: Do you have an example of the output when there is a `HTTPStatusError`?

Just wondering if the print statement gets lost due to the exception being re-raised? If so, we might want to ignore the print and raise a custom exception

rossgray (commented):
Thanks a lot for this, looks really good. Just a small question about what the user sees in the case of an error. I can always take a look at it though if you're away for a bit.

pipeline/cloud/http.py
plutopulp: No worries, currently it looks like:
```shell
pipeline edit pointer betauser/add-lol:v1 --source pipeline_invalid_id
Pipeline 18:14:19 - [INFO]: Editing pointer (betauser/add-lol:v1)
Pipeline 18:14:19 - [ERROR]: {'message': 'Object pipeline=pipeline_invalid_id not found.', 'object_type': 'pipeline', 'lookup_value': 'pipeline_invalid_id', 'x-correlation-id': '2cf8cd07-aa27-400a-adc5-10daf4fbf827'}
Traceback (most recent call last):
  File "/home/ubuntu/code/pipeline/pipeline/console/__init__.py", line 64, in _run
    execute_cli(parser)
  File "/home/ubuntu/code/pipeline/pipeline/console/__init__.py", line 54, in execute_cli
    selected_func(parsed_args)
  File "/home/ubuntu/code/pipeline/pipeline/console/targets/pointers.py", line 218, in _edit_pointer
    http.patch(
  File "/home/ubuntu/code/pipeline/pipeline/cloud/http.py", line 36, in wrapper
    response.raise_for_status()
  File "/home/ubuntu/code/pipeline/.venv/lib/python3.10/site-packages/httpx/_models.py", line 749, in raise_for_status
    raise HTTPStatusError(message, request=request, response=self)
httpx.HTTPStatusError: Client error '404 Not Found' for url 'http://localhost:5025/v4/pointers/betauser/add-lol:v1'
For more information check: https://httpstatuses.com/404
```
But I like the idea of raising a custom exception instead of the print.

rossgray: yeah, I think the error message gets a bit lost in the traceback

plutopulp: So have tried a few things to hide the traceback, using sys/traceback modules, overriding the str method of the custom exception, but no luck.

I think the current way it is displayed in the PR is not the end of the world.

rossgray (approved):

PR #367, opened 2024-01-03
pipeline docker image tweak

plutopulp:
This allows catalyst users to defining their pipeline image without having to prepend by username.
Basically creates a new image name from the response from the API resolved with username, and pushes that to the registry.

### DIFF
--- a/pipeline/console/container/__init__.py
+++ b/pipeline/console/container/__init__.py
@@ -317,7 +317,8 @@ def _push_container(namespace: Namespace):
             raise ValueError("No upload token found")
         if true_pipeline_name is None:
             # will fail unless they use their correct username in the pipeline name
-            # just makes this deployment of pipeline backwards compatible with older catalyst
+            # just makes this deployment of pipeline backwards compatible 
+            # with older catalyst
             true_pipeline_name = pipeline_name
 
         # Login to upload registry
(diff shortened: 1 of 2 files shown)


pipeline/console/container/__init__.py
```diff
@@ -310,9 +311,15 @@ def _push_container(namespace: Namespace):
         )
         start_upload_dict = start_upload_response.json()
         upload_token = start_upload_dict.get("bearer", None)
+        true_pipeline_name = start_upload_dict.get("pipeline_name")
 
         if upload_token is None:
             raise ValueError("No upload token found")
+        if true_pipeline_name is None:
```
rossgray, line 318: still not quite sure in which instances `true_pipeline_name` will be `None`. Is it just until the latest version of catalyst is deployed?

rossgray: if so, perhaps we don't need this check and can assume it's always present

MPCherry: ye its just while we don't have the latest catalyst deployed but people are building and pushing new pipelines. we can remove it if we don't care, but we should just make sure to deploy both changes relatively soon to each other.

rossgray (approved):

PR #374, opened 2024-01-04
Support startup logs by pointer or pipeline id

rossgray:


MPCherry (commented):
LGTM, I think in general though now we're just using `pipeline` when we refer to pipeline id or pointer, e.g. in runs.

MPCherry (approved):

PR #375, opened 2024-01-05
Allow input exceptions to show in run outputs

MPCherry:
For now include run input exceptions in a run's errors. Previously this error would get swallowed and be invisible to the user.

### DIFF
--- a/examples/run_example.py
+++ b/examples/run_example.py
@@ -4,14 +4,13 @@ from pipeline.cloud.pipelines import run_pipeline
 
 start_time = datetime.datetime.now()
 
-pointer = "betauser/simple:v1"
+pointer = "test:v1"
 
-result = run_pipeline(pointer, "a")
+result = run_pipeline(pointer, 1)
 
 end_time = datetime.datetime.now()
 
 total_time = (end_time - start_time).total_seconds()
 
-print(result.id)
 print(f"Total time taken: {total_time}, result: {result.outputs_formatted()}")
-print(f"Total time taken: {total_time}, result: {result.error}")
+print(f"Error: {result.error}")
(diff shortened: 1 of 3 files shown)


examples/run_example.py
```diff
@@ -13,3 +13,4 @@
 total_time = (end_time - start_time).total_seconds()
 
 print(f"Total time taken: {total_time}, result: {result.outputs_formatted()}")
+print(f"Error: {result.error}")
```
rossgray, line 16: I personally use this for printing errors:
```suggestion
if result.error:
    print(f"Error: {result.error.json()}")
else:
    print(
        f"Total time taken: {total_time}\n"
        f"Run ID: {result.id}\n"
        f"Result: {result.outputs_formatted()}"
    )
```

pipeline/container/manager.py
```diff
@@ -219,8 +219,8 @@ def run(
     ) -> t.Any:
         with logger.contextualize(run_id=run_id):
             logger.info("Running pipeline")
-            args = self._parse_inputs(input_data, self.pipeline)
             try:
+                args = self._parse_inputs(input_data, self.pipeline)
```
rossgray, line 223: I think this line could have stayed where it was before?

Otherwise all the `RunInputException`s will be caught and re-raised as `RunnableErrors`.
I guess you can test this though

rossgray (changes requested):

pipeline/container/manager.py
MPCherry: I think this was causing the original issue which is that an input parsing error results in a `500`, and no logs for the run that the user can see. Discussed with Yvan maybe returning a proper HTTP error, we've made a ticket for it here: [PC-952](https://linear.app/mystic-ai/issue/PC-952/pipeline-container-run-error-refinements)

MPCherry: Nvm, as pointed out, we check the exception type and will specifically return a 400 for a run input exception.

rossgray (approved):

PR #378, opened 2024-01-10
Point container pipeline version to latest release

plutopulp:
Change default `pipeline-ai`  version installed in container to be latest release.


MPCherry (approved):
Thanks!

PR #384, opened 2024-01-11
Improve error handling

rossgray:
# Pull request outline

Currently when there are no resources available the error message we return is a bit verbose and could be a bit worrying for users (the default error message from httpx tells users to go to https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503 for more info, which is a bit confusing).

This PR introduces 2 custom exceptions, `APIError` for general HTTP errors and `NoResourcesAvailable` for when a pipeline has not scaled up. This makes the error messaging nicer and also allows the user to use these exceptions in their own code.

## Checklist:
- [x] Version bumped

### DIFF
--- a/pipeline/cloud/http.py
+++ b/pipeline/cloud/http.py
@@ -2,10 +2,9 @@ import functools
 import io
 import os
 import typing as t
+from json.decoder import JSONDecodeError
 
 import httpx
-
-# from httpx import _GeneratorContextManager
 import requests
 from requests_toolbelt import MultipartEncoder, MultipartEncoderMonitor
 from tqdm import tqdm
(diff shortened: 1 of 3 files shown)


pipeline/cloud/http.py
```diff
@@ -17,29 +16,55 @@
...
-        if not isinstance(detail, dict):
-            detail = {"detail": detail}
-    except (TypeError, KeyError):
-        return {"message": "Something went wrong.", "response_json": e.response.json()}
-    return {**detail, "request_id": e.response.headers.get("x-correlation-id")}
+        response.raise_for_status()
+    except httpx.HTTPStatusError as e:
+        error = APIError.from_response(e.response)
+        _print(str(error), level="ERROR")
+        raise error from None
```
plutopulp, line 59: Ah nice so will this suppress the traceback from the `HTTPStatusError`?

plutopulp (approved):
This is great thanks for this @rossgray

pipeline/cloud/http.py
rossgray: yeah, exactly. Only just learnt this :)

PR #388, opened 2024-01-17
Suport for Readme file

hirovi:
# Pull request outline

## Checklist:
- [ ] Docs updated
- [ ] Tests written
- [ ] Version bumped

Added:
- A list of new things

Changed:
- A list of things changed

Removed:
- A list of things removed

## Related issues:
_none_

___

### DIFF
--- a/pipeline/console/container/__init__.py
+++ b/pipeline/console/container/__init__.py
@@ -279,6 +279,11 @@ def _push_container(namespace: Namespace):
 
     pipeline_config = PipelineConfig.parse_obj(pipeline_config_yaml)
 
+    # Read markdown file, transform to string, and put it back in .yaml
+    if pipeline_config.readme is not None:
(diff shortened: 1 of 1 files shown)


rossgray (commented):
I thought `pipeline_config.readme` was usually inline markdown? If we want to support a path to a file then we should probably add in a check to see if `pipeline_config.readme` is a path first before trying to load it. @plutopulp implemented this though, so might be good to confirm with him

plutopulp:
> I thought `pipeline_config.readme` was usually inline markdown? If we want to support a path to a file then we should probably add in a check to see if `pipeline_config.readme` is a path first before trying to load it. @plutopulp implemented this though, so might be good to confirm with him

Yeah guess it would be a good idea to handle that 👍

plutopulp (approved):

PR #394, opened 2024-01-26
prompt user for pipeline container up before pushing

plutopulp:


pipeline/console/container/__init__.py
```diff
@@ -260,6 +260,26 @@ def _build_container(namespace: Namespace):
...
+            "Have you checked your pipeline works locally by running 'pipeline container up'? (yes/no): "
+        )
+        .strip()
+        .lower()
+    )
+    if user_input == "yes":
+        return
+    elif user_input == "no":
+        _print(
+            "Please run the command 'pipeline container up' before continuing. This will help ensure your pipeline works as expected before uploading it to Mystic."
```
MPCherry, line 275: Think we should maybe say something like "We recommend trying `pipeline container up`", since some people won't have the GPU hardware to do it and may think its a requirement.

pipeline/console/container/__init__.py
```diff
@@ -260,6 +260,26 @@ def _build_container(namespace: Namespace):
...
+        )
+        sys.exit()
+    else:
+        _print("Invalid input. Please answer 'yes' or 'no'.")
```
MPCherry, line 279: Maybe just showing the prompt is enough without making them type out yes? Also think we should include a command line option to skip the prompt

rossgray: yeah, I would agree. I was wondering if there's a way we can detect whether they've ran `pipeline container up` and then displaying a warning if they haven't. (Only way I can really think of doing it would be creating a temporary file or saving some kind of marker somewhere)

plutopulp: Cool thanks for the comments, have updated accordingly.
@rossgray Yeah I did wonder about that but considered that getting something simple out to begin with would do.

paulcjh (changes requested):
I left a comment on the Linear ticket for this

paulcjh:
Also, this needs updating to the latest version

PR #399, opened 2024-01-30
Improve error handling in Pipeline

MPCherry:
Fills a gap in our handling of pipeline loading errors, previously we handled failed startups nicely by allowing the pipeline to run but returning a 500 for state. Similarly runs would automatically be redirected with a failed run.

We simply implement a similar idea but for the __init__ of the pipeline. This gives us the following for different cases:

Pipeline failed to load:
```
curl -X POST http://localhost:5005/v4/runs -H 'Authorization: Bearer token' \                                                                        [15:33:08]
--data '{"pipeline": "pipeline_bb16e2b2008a4acab147bfe275608ebc", "inputs": [{"type": "integer", "value": 5}], "async_run": false}' \
--header 'Content-Type: application/json' \

{"inputs":null,"outputs":null,"error":{"type":"startup_error","message":"Pipeline failed to load","traceback":"Traceback (most recent call last):\n  File \"/usr/local/lib/python3.10/site-packages/pipeline/container/manager.py\", line 53, in __init__\n    self.pipeline_module = importlib.import_module(self.pipeline_module_str)\n  File \"/usr/local/lib/python3.10/importlib/__init__.py\", line 126, in import_module\n    return _bootstrap._gcd_import(name[level:], package, level)\n  File \"<frozen importlib._bootstrap>\", line 1050, in _gcd_import\n  File \"<frozen importlib._bootstrap>\", line 1027, in _find_and_load\n  File \"<frozen importlib._bootstrap>\", line 1006, in _find_and_load_unlocked\n  File \"<frozen importlib._bootstrap>\", line 688, in _load_unlocked\n  File \"<frozen importlib._bootstrap_external>\", line 883, in exec_module\n  File \"<frozen importlib._bootstrap>\", line 241, in _call_with_frames_removed\n  File \"/app/new_pipeline.py\", line 38, in <module>\n    output_var = my_model.pedict(input_var)\nAttributeError: 'MyModelClass' object has no attribute 'pedict'. Did you mean: 'predict'?\n"},"id":"run_1a900d9ad71b4c38850a56784074253b","created_at":1706628804.568872,"updated_at":1706628804.568872,"pipeline_id":"pipeline_bb16e2b2008a4acab147bfe275608ebc","state":"failed"}%   
```

Startup failure:
```
ubuntu:test-pl/ (main✗) $ curl -X POST http://localhost:5005/v4/runs -H 'Authorization: Bearer token' \                                                                        [15:34:25]
--data '{"pipeline": "pipeline_401108dded7040abb34d91dfd1e7fe2c", "inputs": [{"type": "integer", "value": 5}], "async_run": false}' \
--header 'Content-Type: application/json' \

{"inputs":null,"outputs":null,"error":{"type":"startup_error","message":"Pipeline failed to startup","traceback":"Traceback (most recent call last):\n  File \"/usr/local/lib/python3.10/site-packages/pipeline/container/manager.py\", line 84, in startup\n    self.pipeline._startup()\n  File \"/usr/local/lib/python3.10/site-packages/pipeline/objects/graph.py\", line 617, in _startup\n    node_function.function(node_function.class_instance, *function_inputs)\n  File \"/app/new_pipeline.py\", line 12, in load\n    raise Exception(\"startup died\")\nException: startup died\n"},"id":"run_a12991fe99b6471885f01980ca41d924","created_at":1706628903.616983,"updated_at":1706628903.616983,"pipeline_id":"pipeline_401108dded7040abb34d91dfd1e7fe2c","state":"failed"}%  
```

Also, have moved the loading state to the init of the pipeline. This will serve the user much better - they are less interested in whether the pipeline is loading vs starting up than whether anything is loading or starting up. This will give a bigger window in which we can make it clear that the pipeline is starting up.

In general I think we have previously differentiated the loading of the pipeline and the startup of the pipeline, but chances are users will put code in different places, sometimes they'll download weights outside the startup function etc., so we should probably treat them more as the same state, as long as users can get a stack trace they'll be able to figure out the issue.

### DIFF
--- a/pipeline/cloud/schemas/pipelines.py
+++ b/pipeline/cloud/schemas/pipelines.py
@@ -117,6 +117,7 @@ class PipelineListPagination(pagination.Pagination):
     order_by: OrderBy
     order: pagination.Order
 
+
 class PipelineDeploymentStatus(str, Enum):
     not_deployed = "not_deployed"
     deploying = "deploying"
--- a/pipeline/cloud/schemas/pipelines.py
+++ b/pipeline/cloud/schemas/pipelines.py
@@ -88,14 +88,6 @@ class PipelineGet(Pipeline):
     extras: t.Optional[dict]
 
 
-class PipelineDeploymentStatus(str, Enum):
-    not_deployed = "not_deployed"
-    deploying = "deploying"
-    deployed = "deployed"
-    failed = "failed"
-    deleting = "deleting"
-    deleted = "deleted"
-
 class PipelinePatch(BaseModel):
     input_variables: t.Optional[t.List[IOVariable]]
     output_variables: t.Optional[t.List[IOVariable]]
@@ -108,8 +100,13 @@ class PipelinePatch(BaseModel):
 
     extras: t.Optional[dict]
 
-    deployment_status: t.Optional[PipelineDeploymentStatus]
-
+class PipelineDeploymentStatus(str, Enum):
+    not_deployed = "not_deployed"
+    deploying = "deploying"
+    deployed = "deployed"
+    failed = "failed"
+    deleting = "deleting"
+    deleted = "deleted"
 
 class PipelineListPagination(pagination.Pagination):
     class OrderBy(str, Enum):
--- a/pipeline/cloud/schemas/pipelines.py
+++ b/pipeline/cloud/schemas/pipelines.py
@@ -134,6 +134,9 @@ class PipelineState(str, Enum):
     load_failed = "load_failed"
     startup_failed = "startup_failed"
 
+    # backwards compatability
+    failed = "failed"
+
 
 class PipelineContainerState(BaseModel):
     state: PipelineState
--- a/pipeline/cloud/schemas/pipelines.py
+++ b/pipeline/cloud/schemas/pipelines.py
@@ -132,8 +132,7 @@ class PipelineState(str, Enum):
     loading = "loading"
     loaded = "loaded"
     load_failed = "load_failed"
-    # This is really startup_failed
-    failed = "failed"
+    startup_failed = "startup_failed"
 
 
 class PipelineContainerState(BaseModel):
--- a/pipeline/cloud/schemas/pipelines.py
+++ b/pipeline/cloud/schemas/pipelines.py
@@ -100,13 +100,6 @@ class PipelinePatch(BaseModel):
 
     extras: t.Optional[dict]
 
-class PipelineDeploymentStatus(str, Enum):
-    not_deployed = "not_deployed"
-    deploying = "deploying"
-    deployed = "deployed"
-    failed = "failed"
-    deleting = "deleting"
-    deleted = "deleted"
 
 class PipelineListPagination(pagination.Pagination):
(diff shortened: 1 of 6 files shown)


pipeline/cloud/schemas/pipelines.py
```diff
@@ -131,6 +131,8 @@ class PipelineState(str, Enum):
     not_loaded = "not_loaded"
     loading = "loading"
     loaded = "loaded"
+    load_failed = "load_failed"
```
paulcjh, line 134: I think changing this to `startup_failed` as commented is better

pipeline/cloud/schemas/pipelines.py
```diff
@@ -131,6 +131,8 @@ class PipelineState(str, Enum):
...
+    # This is really startup_failed
     failed = "failed"
```
paulcjh, line 136: Under what condition would a pipeline be reported as `failed` and not `startup_failed`

MPCherry: Sorry think this is confusing, the comment is intended for `failed`. Basically `load_failed` happens when we import the model, `failed` is our already in place state for an issue during pipeline startup. Hence I commented that it should really be `startup_failed`. I think we'll probably add some more failure modes here so worth refactoring at a later date.

pipeline/container/manager.py
```diff
@@ -61,25 +57,38 @@ def __init__(self, pipeline_path: str):
...
+            self.pipeline_state = pipeline_schemas.PipelineState.load_failed
+            tb = traceback.format_exc()
+            self.pipeline_state_message = tb
+            return
+
     def startup(self):
+        if self.pipeline_state == pipeline_schemas.PipelineState.load_failed:
+            return
         # add context to enable fetching of startup logs
         with logger.contextualize(pipeline_stage="startup"):
```
rossgray, line 84: maybe we could add similar to the `_load()` function so that we can show the full logs to the user?

pipeline/container/manager.py
```diff
@@ -31,7 +31,7 @@ def _get_url_or_path(input_schema: run_schemas.RunInput) -> str | None:
 
 
 class Manager:
-    def __init__(self, pipeline_path: str):
+    def _load(self, pipeline_path: str):
```
rossgray, line 34: maybe we could add `with logger.contextualize(pipeline_stage="loading"):` below here? see comment below

MPCherry: sounds great

pipeline/container/manager.py
MPCherry: Ye that sounds even better tbh.

rossgray: So we'd have to update this line in pcore to query for `startup|loading` if we want those logs to come through, so perhaps worth adding a ticket for that?

rossgray (approved):
LGTM

rossgray (approved):

pipeline/container/manager.py
MPCherry: done https://linear.app/mystic-ai/issue/PC-998/allow-loading-logs

PR #408, opened 2024-02-14
Fix error reporting on pipeline load

rossgray:
# Pull request outline

Currently any exceptions raised during the pipeline load step are not reported back to the end user. This PR fixes this.

## Checklist:
- [x] Version bumped


MPCherry (approved):
Thanks!

PR #419, opened 2024-03-01
Plutopulp/mini frontend refinements

plutopulp:
Container frontend tweaks:
- Fix file downloads
- Fix player for video output


adamleithp-mystic (approved):
I like the no-video library, just html5 is great. It looks the same anyways on modern browsers (which i bet most our users use)

PR #422, opened 2024-03-11
PC-1044 Streaming runs

paulcjh:
# Pull request outline

## Checklist:
- [ ] Docs updated
- [ ] Tests written
- [ ] Version bumped

Added:
- A list of new things

Changed:
- A list of things changed

Removed:
- A list of things removed

## Related issues:
_none_

___


pipeline/cloud/schemas/runs.py
```diff
@@ -130,6 +131,11 @@ def from_object(cls, obj: t.Any):
             except (TypeError, OverflowError):
                 return cls.pkl
             return cls.array
+        # Also check if class name == Stream
+        elif hasattr(obj, "iterable") or (
+            hasattr(obj, "__name__") and obj.__name__ == "Stream"
+        ):
```
rossgray, line 137: I'm not too familiar with this code but can we not just check if `isinstance(obj, Stream)`?

pipeline/container/routes/v4/runs.py
```diff
@@ -80,17 +85,96 @@ async def run(
...
+                                run_schemas.RunOutput(
+                                    type=output.type,
+                                    value=None,
+                                    file=output.file,
+                                )
+                            )
+                            streaming_outputs_to_move.append(output)
+
+                    for output in streaming_outputs_to_move:
+                        streaming_outputs.remove(output)
```
rossgray, line 140: I don't quite understand what's going on here. From what I can see, it looks like if we reach the end of our stream we add a `None` to the final outputs in between the static outputs and the next streaming outputs?

rossgray:
@paulcjh I wonder if we should make the streaming endpoint in the container a separate endpoint, as we're doing with the backend API, since previously we've tried to keep `/v4/runs` consistent between the container and pcore?

pipeline/cloud/schemas/runs.py
paulcjh: I think I did this because your suggestion would cause a circular dependency

rossgray:
the only other thought I had (and maybe we can always do this later) was whether it makes sense to return the full run result schema every time, or just the raw outputs? Since I have a feeling all of the serialising JSON in and out of Pydantic objects may add a bit of overhead. We could always add a separate endpoint for that later though. Maybe also an endpoint where the results are streamed directly from the resource too without going via catalyst and pcore

pipeline/cloud/schemas/runs.py
rossgray: ah ok. Moving the `Stream` class into a separate module may help

plutopulp (approved):
Nice to get this in! We've checked over this a fair few times but had a last look and LGTM

PR #435, opened 2024-03-18
Fix how stream outputs are returned

rossgray:
# Pull request outline

This fixes ordering of outputs in stream endpoint when we have a combination of stream and static outputs and stream outputs return different amounts of data.

## Checklist:
- [x] Tests written


plutopulp (approved):
Nice and simple! LGTM

PR #446, opened 2024-04-09
Plutopulp/pipeline failed state info

plutopulp:
Add `pipeline_state_info` as optional field (non-breaking change) to pipeline get schema, to retrieve information when a pipeline fails to load or startup.


rossgray (approved):
LGTM

PR #448, opened 2024-04-16
PC-1094 Tags on push

paulcjh:
# Pull request outline
This PR adds in the functionality to assign additional pointers/tags to a pipeline when pushing

## Checklist:
- [x] Docs updated
- [x] Version bumped

Added:
- `-p` arg on `pipeline container push`
- `-o` field for force updating a pointer

## Related issues:
_none_

___


pipeline/console/commands.py
```diff
@@ -147,6 +147,18 @@ def container_parser(command_parser: "_SubParsersAction[ArgumentParser]") -> Non
         help="Push a pipeline container.",
     )
     push_parser.set_defaults(func=container._push_container)
+    push_parser.add_argument(
+        "--pointer",
+        "-p",
+        action="append",
+        help="Pointer for the container.",
```
MPCherry, line 154: I'm confused, is this a list or a single value? seems like we assume its a list in the logic but we don't mention that here

paulcjh: you can pass in multiple `-p` args, and I intentionally didn't say it's an array as I thought the user would assume you pass in a list directly to this argument. I've included an example in the docs

rossgray: maybe it should also be `"Pointer for the pipeline." as the pointer is to the pipeline and not the container itself?

PR #450, opened 2024-04-26
change cluster schmea on pipeline get

plutopulp:
Update the cluster response schema tied to a pipeline. Change should be backwards compatible.

### DIFF
--- a/pipeline/cloud/schemas/cluster.py
+++ b/pipeline/cloud/schemas/cluster.py
@@ -6,8 +6,11 @@ class PipelineClusterConfig(BaseModel):
     node_pool: str
 
 
-class PipelineClusterLean(BaseModel):
+class PipelineClusterGetLean(BaseModel):
+    """A lean representation of a cluster when returned from an API call"""
+
     id: str
-    name: str
-    node_pool: str | None = None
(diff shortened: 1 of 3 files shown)


pipeline/cloud/schemas/cluster.py
```diff
@@ -4,3 +4,13 @@
 class PipelineClusterConfig(BaseModel):
     id: str
     node_pool: str
+
+
+class PipelineClusterGetLean(BaseModel):
+    """A lean representation of a cluster when returned from an API call"""
+
+    id: str
+    node_pool: str
```
rossgray, line 13: suppose you could inherit from `PipelineClusterConfig` to avoid copying these fields over?

rossgray (approved):
looks good, just a non-blocking comment

pipeline/cloud/schemas/cluster.py
plutopulp: Yeah I did hesitate doing that. Went against it as `PipelineClusterConfig` is tied to `config.yaml` and it's not clear whether updating that in the future with some fields, whether those fields should also be part of the API response

PR #453, opened 2024-05-01
construct url-safe url when downloading file onto container

plutopulp:
When downloading a file onto the container, we safe-encode the url before downloading the file, e.g. white spaces and other special chars will be encoded etc..
Also, to ensure backward compatibility (for pipelines using older pipeline-ai versions) and to encode any file_urls at the API entry level, we update the RunInput schema to recursively encode any file_url nested within. Note also we ensure that urls are not encoded multiple times.

### DIFF
--- a/pipeline/cloud/schemas/runs.py
+++ b/pipeline/cloud/schemas/runs.py
@@ -5,7 +5,7 @@ from datetime import datetime
 from enum import Enum
 from urllib.parse import quote, unquote
 
-from pydantic import validator
+from pydantic import root_validator, validator
 
 from pipeline.cloud.schemas import BaseModel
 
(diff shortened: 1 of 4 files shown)


MPCherry (approved):
LGTM!

rossgray (approved):
Looks sensible to me. I have to admit, I don't fully understand the whole flow here but what you've done looks correct and like the fact that you've added tests to cover all the different permutations. I wasn't sure whether calling `encode_url` which has a pydantic decorator on it would work outside of pydantic validation but it obviously must do :)

Thanks for fixing this 👍

plutopulp:
> Looks sensible to me. I have to admit, I don't fully understand the whole flow here but what you've done looks correct and like the fact that you've added tests to cover all the different permutations. I wasn't sure whether calling `encode_url` which has a pydantic decorator on it would work outside of pydantic validation but it obviously must do :)
> 
> Thanks for fixing this 👍

Yeah no worries 👍.  I wasn't sure either about calling the pydantic decorated method from outside either, but it worked. Considered moving the encoding logic to a utility function defined outside the pydantic model, but probably fine as is.

PR #454, opened 2024-05-22
Scaling config yaml

MPCherry:
Allow submitting a scaling config name alongside pipeline create.

### DIFF
--- a/pipeline/cloud/schemas/pipelines.py
+++ b/pipeline/cloud/schemas/pipelines.py
@@ -53,6 +53,8 @@ class PipelineCreate(BaseModel):
 
     cluster: PipelineClusterConfig | None = None
 
+    scaling_config: str | None = None
+
     # Additional meta data
     description: t.Optional[str]
     readme: t.Optional[str]
--- a/pipeline/console/container/__init__.py
+++ b/pipeline/console/container/__init__.py
@@ -47,6 +47,7 @@ class PipelineConfig(BaseModel):
     readme: str | None = None
     extras: t.Dict[str, t.Any] | None
     cluster: cluster_schemas.PipelineClusterConfig | None = None
+    scaling_config: str | None = None
 
     class Config:
         extra = "forbid"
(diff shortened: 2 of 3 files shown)


rossgray (commented):
The only question I have is whether `scaling_config` represents a scaling config ID or name? If it's just one or the other, maybe it's worth naming it `scaling_config_id` or `scaling_config_name` so that it's obvious to people adding it to their YAML? Or if we're gonna support both then should be fine to leave as is

MPCherry:
> The only question I have is whether `scaling_config` represents a scaling config ID or name? If it's just one or the other, maybe it's worth naming it `scaling_config_id` or `scaling_config_name` so that it's obvious to people adding it to their YAML? Or if we're gonna support both then should be fine to leave as is

Its a name, I've copied the convention we used in the pipeline patch, but agree its slightly confusing.

rossgray:
> > The only question I have is whether `scaling_config` represents a scaling config ID or name? If it's just one or the other, maybe it's worth naming it `scaling_config_id` or `scaling_config_name` so that it's obvious to people adding it to their YAML? Or if we're gonna support both then should be fine to leave as is
> 
> Its a name, I've copied the convention we used in the pipeline patch, but agree its slightly confusing.

ok, I guess we'll have to document this somewhere anyway, so we can make sure we specify that it needs to be the name and not the ID

rossgray (approved):

MPCherry:
> > > The only question I have is whether `scaling_config` represents a scaling config ID or name? If it's just one or the other, maybe it's worth naming it `scaling_config_id` or `scaling_config_name` so that it's obvious to people adding it to their YAML? Or if we're gonna support both then should be fine to leave as is
> > 
> > 
> > Its a name, I've copied the convention we used in the pipeline patch, but agree its slightly confusing.
> 
> ok, I guess we'll have to document this somewhere anyway, so we can make sure we specify that it needs to be the name and not the ID

Just changed it anyway lol, for the api calls and models themselves we can stick to this convention but the config having name just makes it clearer.

rossgray (approved):
nice!

PR #459, opened 2024-05-30
Plutopulp/safe encode file path frontend container

plutopulp:
# Pull request outline

## Checklist:
- [x] Version bumped

Changed:
- Before uploading a file from the (local) container play UI, we encode the file name. This fixes special character issues when constructing the file path on the container, e.g. white spaces etc..


paulcjh (approved):
LGTM

PR #461, opened 2024-06-03
Fix "Not supported URL scheme http+docker" error

rossgray:
# Pull request outline

There is an issue with the latest version of `requests` which breaks Docker commands with versions of `docker < 7.1.0`. This PR updates both the `docker` and `requests` dependencies to resolve this.

There is also a small fix for error handling when a container fails to start up when using `pipeline container up`.

## Checklist:
- [x] Version bumped


MPCherry (approved):
Thanks!

PR #463, opened 2024-06-07
Reduce Docker image size by excluding pip cache dir

rossgray:
# Pull request outline

We can reduce the resultant pipeline Docker image size by specifying the `--no-cache-dir` option when running `pip install`. When testing locally, this can significantly reduce Docker image sizes (for example it reduces a Stable Diffusion image by almost 3GB). This should lead to faster cold start times.

## Checklist:
- [x] Version bumped


plutopulp (approved):
LGTM, thanks for this @rossgray.

PR #468, opened 2024-07-16
Turbo Registry Integration

MPCherry:
Adds turbo registry integration to the pipeline client.

With this, turbo registry deployments can be enabled by adding a value to extras as so:
```
extras:
  turbo_registry: true
```


pipeline/console/container/push.py
```diff
@@ -86,7 +86,11 @@ def push_container(namespace: Namespace):
 
     docker_client: docker.DockerClient = docker.from_env(timeout=600)
 
-    registry_info = http.get(endpoint="/v4/registry")
+    registry_params = {}
+    if pipeline_config.extras and "turbo_registry" in pipeline_config.extras:
```
rossgray, line 90: think we should implement the same check as above, i.e.
```suggestion
    if pipeline_config.extras and pipeline_config.extras.get("turbo_registry", False):
```

rossgray (commented):
Just one small suggestion, looks good apart from that

pipeline/console/container/push.py
MPCherry: whoops, missed this one

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "pipeline-ai"
-version = "2.3.2"
+version = "2.3.3"
```
rossgray, line 3: I would probably make this a [minor update rather than a patch](https://semver.org/#summary) as you're adding new functionality
```suggestion
version = "2.4.0"
```

rossgray (approved):

PR #469, opened 2024-07-18
Single Layer Turbo Validation

MPCherry:
Adds checks for building and pushing turbo registry images. 

Images are required to be a single layer, customer UX will be improved if we catch these client side and direct them to support.


paulcjh (approved):
yay

PR #475, opened 2024-07-26
Add tests for container routes

rossgray:
# Pull request outline

Add some tests for pipeline container routes, which are currently untested.

In order to get the tests working, I added a `lifespan` event for starting the execution handler loop: https://github.com/mystic-ai/pipeline/pull/475/files#diff-56467ebb52b06ad6784e7b2347f04e2947276d518f0b57f3f239ae45db46cbb3R46-R59

(without this, the tests would hang as the coroutine never terminates, as it runs in an infinite loop).

I made the mistake of updating the version of Black we're using for autoformatting and it resulted in quite a few changes.

I also used this opportunity to remove the CLI commands related to environments, which are no longer used in V4 (these were also failing Black checks so thought it was best to remove them entirely).

I haven't 

## Checklist:
- [x] Tests written
- [x] Version bumped


pipeline/console/targets/environments.py
plutopulp, line 1: Thanks for purging all the environments stuff

plutopulp (approved):
LGTM

PR #476, opened 2024-07-30
Publish current run from `v4/container/state` endpoint

rossgray:
# Pull request outline

As well as publishing the current run, there are a few more changes in here (looking at the commits individually will help review):
- refactor the `execution_handler` slightly
    - move into a run services module
    - it now returns the result and status code rather than just the result
- update a few dependencies (including httpx)
- update the tests
    - needed to use an async client in order to mock certain behaviour (if we don't use an async client we get issues mocking things as they run in a different event loop to the API)  

## Checklist:
- [x] Tests written
- [x] Version bumped

### DIFF
--- a/pipeline/container/manager.py
+++ b/pipeline/container/manager.py
@@ -37,7 +37,7 @@ class Manager:
             pipeline_schemas.PipelineState.not_loaded
         )
         self.pipeline_state_message: str | None = None
-        self.current_run_id: str | None = None
+        self.current_run: str | None = None
         try:
             self._load(pipeline_path)
         except Exception:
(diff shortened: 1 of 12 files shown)


pipeline/container/manager.py
```diff
@@ -242,12 +243,15 @@ def run(
     ) -> t.Any:
         with logger.contextualize(run_id=run_id):
             logger.info("Running pipeline")
-            args = self._parse_inputs(input_data, self.pipeline)
+            self.current_run_id = run_id
             try:
+                args = self._parse_inputs(input_data, self.pipeline)
```
plutopulp, line 248: Good catch 👍 , I guess any errors before from parsing run inputs weren't being caught by `RunInputException`?

rossgray: I think they're being caught a level higher up but other exceptions weren't, and needed to make sure any exception in this method is caught so that the current run is always reset

plutopulp (approved):
LGTM, although maybe another pair of eyes may be good as not super familiar with the code.

PR #477, opened 2024-07-31
Update schemas to support async runs

rossgray:
# Pull request outline

Here we update the run schemas to support async runs in the pipeline container (which are coming soon).

It no longer makes sense to share the run schemas across the main API and the pipeline container since several params are now specific to one or the other, so it is best to separate them to avoid confusion. For example `run_id` is an optional parameter for `ContainerRunCreate` but was inherited by `RunCreate`, even though it is not used by the main API.

Main changes here are adding `async_run` and `callback_url` to `ContainerRunCreate` - these will be needed for async runs. These are optional so shouldn't cause anything to break

I have kept `run_id` in `RunCreate` just for backwards compatibility, in case users are using this unknowingly (we can probably check in the main API if it's being used and deprecate if not)

## Checklist:
- [x] Version bumped

### DIFF
--- a/pipeline/cloud/schemas/runs.py
+++ b/pipeline/cloud/schemas/runs.py
@@ -329,3 +329,6 @@ class RunCreate(BaseModel):
     # flag to determine whether the run will wait for compute resources to be
     # become available if none are currently running the pipeline
     wait_for_resources: bool | None = None
+    # run_id is not used and should not be provided (it is kept here for
+    # backwards compatibility)
+    run_id: str | None = None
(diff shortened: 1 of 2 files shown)


pipeline/cloud/schemas/runs.py
```diff
@@ -265,14 +265,16 @@ class ContainerRunError(BaseModel):
 
 class ContainerRunCreate(BaseModel):
     # run_id is optional since it's just used for attaching logs to a run
-    run_id: t.Optional[str]
-    inputs: t.List[RunInput]
+    run_id: str | None
+    inputs: list[RunInput]
+    async_run: bool = False
+    callback_url: str | None
```
plutopulp, line 271: Is this gonna be a pre-signed url type of thing for the container to publish result to main api or something?

plutopulp (approved):
Yeah looks good, just a wee question about `callback_url`

pipeline/cloud/schemas/runs.py
rossgray: yes exactly

PR #479, opened 2024-08-12
Better reporting of readme generation errors

rossgray:
# Pull request outline

Generating the README can sometimes fail, as it is parsed as a Python template string. Here we add better reporting of errors to make it easier to resolve these.

Example of new error:
```
❯ pipeline container push
Pipeline 10:13:00 - [ERROR]: Failed to generate README: KeyError('hmm')
You may need to escape certain characters, eg change `{ }` to `{{ }}`
```

## Checklist:
- [x] Version bumped


plutopulp (approved):
LGTM

